PR #11, opened 2020-03-27
Ensure no redundant span.kind tag for Zipkin spans

rmfitzpatrick:
Prevents the `span.kind` OT tag used to determine the Zipkin span kind from being sent w/ Zipkin spans.  Also ensures the Zipkin kind is uppercase.

### DIFF
--- a/test/Datadog.Trace.TestHelpers/ZipkinHelpers.cs
+++ b/test/Datadog.Trace.TestHelpers/ZipkinHelpers.cs
@@ -113,10 +113,20 @@ namespace Datadog.Trace.TestHelpers
             if (expected.Tags != null)
             {
                 var actualTags = new Dictionary<string, string>(actual.Tags());
+                var expectedTags = new Dictionary<string, string>(expected.Tags);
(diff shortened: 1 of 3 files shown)


test/Datadog.Trace.TestHelpers/ZipkinHelpers.cs
```diff
@@ -112,7 +112,11 @@ public static void AssertSpanEqual(Span expected, JToken actual)
 
             if (expected.Tags != null)
             {
-                Assert.Equal(expected.Tags, actual.Tags());
+                var actualTags = new Dictionary<string, string>(actual.Tags());
```
johnbley: If the point of the change is to make sure that kind's value is capitalized, shouldn't there be a test for that?  I can't find one in the rest of this file either.

rmfitzpatrick: Good call.  Updated.

rmfitzpatrick:
@johnbley, mind taking another look?

johnbley:
👍

PR #12, opened 2020-04-07
Fix net45 builds

pjanotti:
The implementation in TimeExtensions was depending on a method not available on net45, however, that was only needed in tests so created a TimeHelpers on the test helpers to cover that old extension method that needed to be removed.


src/Datadog.Trace/ExtensionMethods/TimeExtensions.cs
```diff
@@ -22,7 +22,9 @@ public static long ToUnixTimeNanoseconds(this DateTimeOffset dateTimeOffset)
         /// <returns>The number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.</returns>
         public static long ToUnixTimeMicroseconds(this DateTimeOffset dateTimeOffset)
         {
-            return ToUnixTimeNanoseconds(dateTimeOffset) / 1000;
+            const long microsecondsPerMillisecond = 1000;
```
rmfitzpatrick, line 25: Thanks for making this more accurate.  It's my typo, but would you mind correcting the return comment?

PR #13, opened 2020-04-08
Update redis span content w/ OT tags

rmfitzpatrick:
These changes update the existing redis instrumentations to conform to OT conventions.  Also setting the component tag to be the assembly name to better match libraries.


PackageVersionsGeneratorDefinitions.json
```diff
@@ -1,4 +1,5 @@
 [
+  // Modified by SignalFx
```
pjanotti, line 2: I don't think you can have comments in JSON files...

src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs
```diff
@@ -36,10 +36,12 @@ internal static Scope CreateScope(Tracer tracer, string integrationName, string
...
 
+                scope = tracer.StartActive(command ?? OperationName, serviceName: serviceName);
+
                 var span = scope.Span;
-                span.Type = SpanTypes.Redis;
-                span.ResourceName = command;
-                span.SetTag(Tags.RedisRawCommand, rawCommand);
+                span.SetTag(Tags.InstrumentationName, integrationName);
+                span.SetTag(Tags.DbType, SpanTypes.Redis);
+                span.SetTag(Tags.DbStatement, rawCommand.Length > 1024 ? rawCommand.Substring(0, 1024) : rawCommand);
```
pjanotti, line 44: nit: careful hand compiled
```suggestion
                span.SetTag(Tags.DbStatement, rawCommand.Substring(0, Math.Min(rawCommand.Length, 1024)));
```

src/Datadog.Trace.ClrProfiler.Managed/Integrations/ServiceStackRedisIntegration.cs
```diff
@@ -82,7 +84,7 @@ public static class ServiceStackRedisIntegration
 
             using (var scope = RedisHelper.CreateScope(
                 Tracer.Instance,
-                IntegrationName,
+                RedisAssembly,
```
pjanotti, line 87: It looks like this change will break this check https://github.com/signalfx/signalfx-dotnet-tracing/blob/2fb3280c55317d0a9804cb352163fa7ec988ec34/src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs#L16

src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs
```diff
@@ -36,10 +36,12 @@ internal static Scope CreateScope(Tracer tracer, string integrationName, string
                     command = rawCommand;
                 }
 
+                scope = tracer.StartActive(command ?? OperationName, serviceName: serviceName);
+
                 var span = scope.Span;
-                span.Type = SpanTypes.Redis;
```
pjanotti, line 40: For my enlightenment what this and the property below mean for dd? Shouldn't they be kept?

PackageVersionsGeneratorDefinitions.json
rmfitzpatrick: Generally not, but C# seems to support them (and I've not run into issues so far) https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.readcommenthandling?view=netcore-3.1.

Is this causing problems on windows?  We'd need to remove a number of them from samples and other places.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/ServiceStackRedisIntegration.cs
rmfitzpatrick: Good catch, I made a mental note to avoid this but missed the check in the helper.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs
rmfitzpatrick: They aren't OT-specific and aren't used by our product so we've made minor efforts to remove them.  `db.type` is preferable imo.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs
pjanotti: Since this pattern repeats a lot perhaps having a truncate helper seems a good idea. Also, the value 1024 happens in many places of the code it will be good to coalesce but let's do that as a separate PR.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/RedisHelper.cs
rmfitzpatrick: Resource name appears to be akin to operation name, which is where we generally want it since the default operation name is of little value.

PackageVersionsGeneratorDefinitions.json
pjanotti: I see, so it depends on the tooling, if not a problem on the repo then keep it.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/ServiceStackRedisIntegration.cs
rmfitzpatrick: updated

PR #14, opened 2020-04-09
Remove GetValueOrDefault on ES tests

pjanotti:
GetValueOrDefault is not available on some of the older target frameworks, using TryGetValue instead.

Fixes Azure Pipelines for Windows


rmfitzpatrick:
These should be using the extension method: https://github.com/signalfx/signalfx-dotnet-tracing/blob/master/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs#L21.  Can you import and add the string generic type so it sources correctly?  My mistake.

pjanotti:
Due to the use of `System.Collection.Generics` on the files we can't use it as an extension method (it will be ambiguous) so the code will have to use `DictionaryExtensions.GetValueOrDefault`. Let's see how it looks.

PR #19, opened 2020-04-17
Fix race on MockZipkinCollector HttpListener

pjanotti:
Trying a minimal fix for issues that I found running on Windows and I suspect it also to be related to Linux Azure Pipeline not finishing. This is basically a quick workaround on the limited HttpListener type, that said there is no good and simple alternative to it to use in tests.

### DIFF
--- a/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
+++ b/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
@@ -140,7 +140,6 @@ namespace Datadog.Trace.TestHelpers
             {
                 _listenerCts.Cancel();
                 _listener.Stop();
-                _listenerCts.Dispose();
             }
         }
 
--- a/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
+++ b/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
@@ -11,6 +11,7 @@ using System.Net;
 using System.Runtime.Serialization;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using Datadog.Trace.ExtensionMethods;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -22,6 +23,7 @@ namespace Datadog.Trace.TestHelpers
     {
         private readonly HttpListener _listener;
         private readonly Thread _listenerThread;
+        private readonly CancellationTokenSource _listenerCts = CancellationTokenSource();
 
         public MockZipkinCollector(int port = 9080, int retries = 5)
         {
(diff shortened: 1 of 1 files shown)


rmfitzpatrick:
Would you be able to describe the race condition?  It's not immediately clear to me what your changes are fixing.

test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
```diff
@@ -167,11 +173,14 @@ protected virtual void OnRequestDeserialized(IList<IMockSpan> trace)
 
         private void HandleHttpRequests()
         {
-            while (_listener.IsListening)
```
pjanotti, line 170: Here it was possible to read `true` but on the call below to `_listener.GetContext()` the listener was already closed causing an object already disposed exception (encountered in practice).

test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
```diff
@@ -167,11 +173,14 @@ protected virtual void OnRequestDeserialized(IList<IMockSpan> trace)
...
+            while (true)
             {
                 try
                 {
-                    var ctx = _listener.GetContext();
+                    var getCtxTask = Task.Run(() => _listener.GetContext());
```
pjanotti, line 180: `GetContext()` on Windows terminates when the `_listener` is closed but not on Linux so this could block forever.

test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
```diff
@@ -197,10 +201,17 @@ private void HandleHttpRequests()
                     ctx.Response.OutputStream.Write(buffer, 0, buffer.Length);
                     ctx.Response.Close();
                 }
-                catch (HttpListenerException)
+                catch (Exception ex) when (ex is OperationCanceledException || ex is AggregateException)
                 {
-                    // listener was stopped,
-                    // ignore to let the loop end and the method return
+                    lock (_listener)
```
pjanotti, line 206: The lock is needed here in conjunction with the one on the dispose to ensure that the flag behind `IsListening` is the latest value, otherwise, the `_listener` can be disposed but the check here may read `IsListening` still as true (encountered in practice while testing the fix).

pjanotti:
> Would you be able to describe the race condition? It's not immediately clear to me what your changes are fixing.

Added a few comments, let me know if that is not clear.

test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
pjanotti: Notice that there is a potential thread leak here but that should be fine for test code (the endpoint is freed by the dispose).

PR #22, opened 2020-04-20
Update ZipkinContent according to SendTracesAsync changes

pjanotti:
The signature of the method was changed upstream and initially only minimal change was made to accomodate the new signature. This change fully updates the stack according to the new signature.

### DIFF
--- a/src/Datadog.Trace/Agent/ZipkinContent.cs
+++ b/src/Datadog.Trace/Agent/ZipkinContent.cs
@@ -12,21 +12,20 @@ namespace Datadog.Trace.Agent
 {
     internal class ZipkinContent : HttpContent
     {
-        private readonly ZipkinSerializer serializer = new ZipkinSerializer();
+        private readonly ZipkinSerializer _serializer = new ZipkinSerializer();
+        private readonly Span[][] _spans;
 
-        public ZipkinContent(Span[][] value)
+        public ZipkinContent(Span[][] spans)
         {
-            Value = value;
+            _spans = spans;
             Headers.ContentType = new MediaTypeHeaderValue("application/json");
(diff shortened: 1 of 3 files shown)


src/Datadog.Trace/Agent/ZipkinContent.cs
```diff
@@ -10,24 +10,23 @@
 
 namespace Datadog.Trace.Agent
 {
-    internal class ZipkinContent<T> : HttpContent
```
pjanotti, line 13: @rmfitzpatrick let me know if for some reason that I missed we need this generic here.

src/Datadog.Trace/Agent/ZipkinContent.cs
```diff
@@ -10,24 +10,23 @@
...
 
-        public ZipkinContent(T value)
+        public ZipkinContent(Span[][] value)
         {
             Value = value;
             Headers.ContentType = new MediaTypeHeaderValue("application/json");
         }
 
-        public T Value { get; }
+        public Span[][] Value { get; }
```
pjanotti, line 23: Hum it seems that could be a private field...

src/Datadog.Trace/Agent/ZipkinContent.cs
rmfitzpatrick: It was taken from the `MsgPackContent` basis, which doesn't seem necessary.

src/Datadog.Trace/Agent/ZipkinContent.cs
rmfitzpatrick: same story as generic parameter

PR #23, opened 2020-04-20
Fix new race in MockZipkinCollector

pjanotti:
This race was introduced on the previous PR since HttpListenerException can still be raised if the _listener is disposed of but the code still tries to write to the output stream.

### DIFF
--- a/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
+++ b/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs
@@ -201,7 +201,7 @@ namespace Datadog.Trace.TestHelpers
                     ctx.Response.OutputStream.Write(buffer, 0, buffer.Length);
                     ctx.Response.Close();
                 }
-                catch (Exception ex) when (ex is OperationCanceledException || ex is AggregateException)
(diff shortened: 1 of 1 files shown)


pjanotti:
```terminal
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.Net.HttpListenerException (0x80131620): Unable to write data to the transport connection: Cannot access a disposed object.
Object name: 'System.Net.Sockets.Socket'..
   at System.Net.HttpResponseStream.InternalWrite(Byte[] buffer, Int32 offset, Int32 count)
   at System.Net.HttpResponseStream.WriteCore(Byte[] buffer, Int32 offset, Int32 size)
   at System.Net.HttpResponseStream.Write(Byte[] buffer, Int32 offset, Int32 size)
   at Datadog.Trace.TestHelpers.MockZipkinCollector.HandleHttpRequests() in /project/test/Datadog.Trace.TestHelpers/MockZipkinCollector.cs:line 201
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ThreadHelper.ThreadStart()
```

PR #25, opened 2020-04-21
Adopt ADO.NET updates and add SQL sanitization 

rmfitzpatrick:
Verifies upstream ADO.NET and Npqsql updates as well as ports the sql sanitization regexes and tests* from https://github.com/signalfx/signalfx-java-tracing/commit/123b8104d361f07d87a0caf445479a07e203bf5a.  I'm providing an opt-out mechanism as well for services that aren't dealing with PII.

Also bumps up the graphql timeout to reduce flake until we source the most recent improvements.

### DIFF
--- a/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
+++ b/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
@@ -45,7 +45,11 @@ namespace Datadog.Trace.ExtensionMethods
                 }
             }
 
-            span.SetTag(Tags.DbStatement, statement);
+            if (!string.IsNullOrEmpty(statement))
+            {
+                span.SetTag(Tags.DbStatement, statement);
+            }
+
             span.SetTag(Tags.SpanKind, SpanKinds.Client);
 
             // parse the connection string
--- a/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
+++ b/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
@@ -30,12 +30,19 @@ namespace Datadog.Trace.ExtensionMethods
         /// </summary>
         /// <param name="span">The span to add the tags to.</param>
         /// <param name="command">The db command to get tags values from.</param>
-        /// <param name="statement">The db statement to use over command.CommandText.</param>
-        public static void AddTagsFromDbCommand(this Span span, IDbCommand command, string statement = "")
+        /// <param name="statement">The db statement to use over command.CommandText. Will not be truncated or sanitized.</param>
+        /// <param name="sanitize">Whether to sanitize tagged statements.</param>
+        public static void AddTagsFromDbCommand(this Span span, IDbCommand command, string statement = "", bool sanitize = true)
         {
             if (string.IsNullOrEmpty(statement))
             {
-                statement = command.CommandText.Length > 1024 ? command.CommandText.Substring(0, 1024) : command.CommandText;
+                statement = command.CommandText ?? string.Empty;
+                statement = statement.Truncate();
+
+                if (sanitize)
+                {
+                    statement = statement.SanitizeSqlStatement();
+                }
             }
 
             span.SetTag(Tags.DbStatement, statement);
--- a/src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
+++ b/src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
@@ -1,9 +1,24 @@
 using System;
+using System.Text.RegularExpressions;
 
 namespace Datadog.Trace.ExtensionMethods
 {
     internal static class StringExtensions
     {
+        private static string hex = "(0x[0-9a-fA-F]+)";
+
+        private static string numericLiteral = @"((?<![\p{L}\p{C}_0-9])[+.-]*[0-9]+[0-9xEe.+-]*)";
+
+        private static string singleQuoted = "('((?:''|[^'])*)')";
+
+        private static string doubleQuoted = "(\"((?:\"\"|[^\"])*)\")";
+
+        private static string replacement = "?";
+
+        private static string sqlNormalizeString = string.Join("|", new string[] { hex, numericLiteral, singleQuoted, doubleQuoted });
+
+        private static Regex sqlNormalizePattern = new Regex(sqlNormalizeString, RegexOptions.Compiled);
+
         /// <summary>
         /// Removes the trailing occurrence of a substring from the current string.
         /// </summary>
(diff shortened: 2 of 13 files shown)


test/Datadog.Trace.ClrProfiler.IntegrationTests/GraphQLTests.cs
```diff
@@ -95,7 +95,7 @@ public void SubmitsTraces()
                 process.BeginErrorReadLine();
 
                 // wait for server to start
-                wh.WaitOne(5000);
+                wh.WaitOne(10000);
```
pjanotti, line 98: 👍

test/Datadog.Trace.Tests/ExtensionMethods/StringExtensionsTests.cs
```diff
@@ -20,5 +21,52 @@ public void TrimEnd(string original, string suffix, string expected)
             string actual = original.TrimEnd(suffix, StringComparison.Ordinal);
             Assert.Equal(expected, actual);
         }
+
```
pjanotti, line 24: It is a trivial method but a simple test for `Truncate` will be good.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -31,11 +31,19 @@ public static void SetTraceSamplingPriority(this Span span, SamplingPriority sam
...
+        /// <param name="sanitize">Whether to sanitize tagged statements.</param>
+        public static void AddTagsFromDbCommand(this Span span, IDbCommand command, string statement = "", bool sanitize = true)
         {
             if (string.IsNullOrEmpty(statement))
             {
-                statement = command.CommandText.Length > 1024 ? command.CommandText.Substring(0, 1024) : command.CommandText;
+                statement = command.CommandText;
+            }
+
+            statement = statement.Truncate();
```
pjanotti, line 42: If `command.CommandText` can be `null` the call to `Truncate()` will raise `NullReferenceException`.

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
```diff
@@ -46,5 +61,25 @@ public static string TrimEnd(this string value, string suffix, StringComparison
...
         }
+
+        /// <summary>
+        /// Truncates a <see cref="string"/> to 1024 characters, if necessary.
+        /// </summary>
+        /// <param name="value">The string to convert.</param>
+        /// <returns>The truncated string</returns>
+        public static string Truncate(this string value)
+        {
+          return value.Length > 1024 ? value.Substring(0, 1024) : value;
```
pjanotti, line 72: Decide what to do if `value` is `null`. let the exception be raised or return a place holder?

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
```diff
@@ -46,5 +61,25 @@ public static string TrimEnd(this string value, string suffix, StringComparison
...
+        }
+
+        /// <summary>
+        /// Sanitize all values from sql statements.
+        /// </summary>
+        /// <param name="value">The string to sanitize.</param>
+        /// <returns>The santized string</returns>
+        public static string SanitizeSqlStatement(this string value)
+        {
+            return sqlNormalizePattern.Replace(value, replacement);
```
pjanotti, line 82: What happens if `value` is `null`?

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
rmfitzpatrick: Docs say the default is an empty string, but I added an assurance.

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
rmfitzpatrick: Following the pattern established by the other extension methods.

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
rmfitzpatrick: same as other item

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -30,12 +30,19 @@ public static void SetTraceSamplingPriority(this Span span, SamplingPriority sam
...
-        /// <param name="statement">The db statement to use over command.CommandText.</param>
-        public static void AddTagsFromDbCommand(this Span span, IDbCommand command, string statement = "")
+        /// <param name="statement">The db statement to use over command.CommandText. Will not be truncated or sanitized.</param>
+        /// <param name="sanitize">Whether to sanitize tagged statements.</param>
+        public static void AddTagsFromDbCommand(this Span span, IDbCommand command, string statement = "", bool sanitize = true)
         {
             if (string.IsNullOrEmpty(statement))
             {
-                statement = command.CommandText.Length > 1024 ? command.CommandText.Substring(0, 1024) : command.CommandText;
+                statement = command.CommandText ?? string.Empty;
```
pjanotti, line 39: It seems that the idea was to close the `if` after this line so truncate and sanitize happen always.

test/Datadog.Trace.Tests/ExtensionMethods/StringExtensionsTests.cs
```diff
@@ -20,5 +21,62 @@ public void TrimEnd(string original, string suffix, string expected)
...
+        [InlineData("SELECT * FROM TABLE WHERE FIELD = \"a single ' singlequote inside\"", "SELECT * FROM TABLE WHERE FIELD = ?")]
+        public void SqlStatementsAreSanitized(string original, string expected)
+        {
+            string actual = original.SanitizeSqlStatement();
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
+        public void TruncateExtension()
+        {
```
pjanotti, line 74: nice Azure Pipeline feature :) shows the errors on the source.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
pjanotti: Perhaps a test for it too?

rmfitzpatrick: I updated it so that if `statement` is provided it won't be truncated or sanitized.  The ScopeFactory is only using this path because it needs to confirm that the potential parent span's statement doesn't match its truncated and sanitized form, which would lead to redundant calls: https://github.com/signalfx/signalfx-dotnet-tracing/pull/25/files#diff-80868d7b192ea9434ed9d2c3a35b7b78R119

rmfitzpatrick: test added

pjanotti: I think that it is worth a comment, code readers may think it is a bug.

PR #26, opened 2020-04-23
Migrate error tags to sfx.error.*

rmfitzpatrick:
Adopts error analyzer tag names* similar to previous java and node updates.

### DIFF
--- a/src/Datadog.Trace/Tags.cs
+++ b/src/Datadog.Trace/Tags.cs
@@ -68,17 +68,17 @@ namespace Datadog.Trace
         /// <summary>
         /// The error message of an exception
         /// </summary>
-        public const string ErrorMsg = "error.msg";
+        public const string ErrorMsg = "sfx.error.message";
 
         /// <summary>
-        /// The type of an exception
+        /// The kind of exception
         /// </summary>
-        public const string ErrorType = "error.type";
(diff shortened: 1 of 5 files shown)


src/Datadog.Trace/Tags.cs
```diff
@@ -68,17 +68,17 @@ public static class Tags
...
         /// </summary>
-        public const string ErrorMsg = "error.msg";
+        public const string ErrorMsg = "sfx.error.message";
 
         /// <summary>
-        /// The type of an exception
+        /// The kind of exception
         /// </summary>
-        public const string ErrorType = "error.type";
+        public const string ErrorKind = "sfx.error.kind";
```
owais, line 76: Should `ErrorType` continue to exist as an alias to `ErrorKind`? Is it possible that users might be referring this variable somewhere in their code?

rmfitzpatrick: Though not explicitly documented, custom instrumentation tags should be pulled from https://github.com/opentracing/opentracing-csharp/blob/master/src/OpenTracing/Tag/Tags.cs#L63 which is unchanged.  This is a good point across all our offerings though, as the preference has been to advise uses not to use internal resources no matter their access modifiers as they're change prone.  @brianashby-sfx (fyi)

PR #35, opened 2020-05-18
Update GraphQL span content

rmfitzpatrick:
These changes include providing more informative operation names, truncating tagged queries, and adopting Zipkin encoding in the corresponding integration test suite.

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Datadog.Trace.ClrProfiler.Emit;
 using Datadog.Trace.ClrProfiler.Helpers;
+using Datadog.Trace.ExtensionMethods;
 using Datadog.Trace.Logging;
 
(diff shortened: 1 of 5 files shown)


src/Datadog.Trace.ClrProfiler.Managed/Integrations/GraphQLIntegration.cs
```diff
@@ -262,7 +261,7 @@ private static Scope CreateScopeFromValidate(object document)
                 scope = tracer.StartActive(ValidateOperationName, serviceName: serviceName);
                 var span = scope.Span;
                 DecorateSpan(span);
-                span.SetTag(Tags.GraphQLSource, source);
+                span.SetTag(Tags.GraphQLSource, source.Truncate());
```
pjanotti, line 264: What about moving the truncation to `SetTag`? Better as a separate PR/discussion.

rmfitzpatrick: I think we definitely need that to avoid sending arbitrarily long tag values (adding `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` like we do in java), but at the instrumentation level the 1024 char limit makes more sense per value instance imo.

PR #38, opened 2020-06-09
Add HttpClient support

pjanotti:
This was already supported by the code this change just put it closer to OpenTelemetry current specs for HTTP spans by changing the span name and also adding `http.status_text` for failed status codes.

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
@@ -201,7 +201,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                         scope.Span.SetTag(Tags.HttpStatusCode, ((int)response.StatusCode).ToString());
                         if (!response.IsSuccessStatusCode && !string.IsNullOrWhiteSpace(response.ReasonPhrase))
                         {
-                            scope.Span.SetTag(Tags.HttpStatusText, response.ReasonPhrase);
+                            scope.Span.SetTag(Tags.HttpStatusText, response.ReasonPhrase.Truncate());
                         }
                     }
 
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
@@ -172,7 +172,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
             HttpRequestMessage request,
             CancellationToken cancellationToken)
         {
-            if (!(handler is HttpClientHandler || reportedType.FullName.Equals("System.Net.Http.SocketsHttpHandler", StringComparison.OrdinalIgnoreCase)) ||
+            if (!(handler is HttpClientHandler || "System.Net.Http.SocketsHttpHandler".Equals(reportedType.FullName, StringComparison.OrdinalIgnoreCase)) ||
                 !IsTracingEnabled(request))
             {
                 // skip instrumentation
@@ -195,8 +195,15 @@ namespace Datadog.Trace.ClrProfiler.Integrations
 
                     HttpResponseMessage response = await sendAsync(handler, request, cancellationToken).ConfigureAwait(false);
 
-                    // this tag can only be set after the response is returned
-                    scope?.Span.SetTag(Tags.HttpStatusCode, ((int)response.StatusCode).ToString());
+                    if (scope != null)
+                    {
+                        // this tag can only be set after the response is returned
+                        scope.Span.SetTag(Tags.HttpStatusCode, ((int)response.StatusCode).ToString());
+                        if (!response.IsSuccessStatusCode && !string.IsNullOrWhiteSpace(response.ReasonPhrase))
+                        {
+                            scope.Span.SetTag(Tags.HttpStatusText, response.ReasonPhrase);
+                        }
+                    }
 
                     return response;
                 }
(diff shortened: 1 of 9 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -472,7 +472,7 @@ private static void RegisterGlobalTracer(Tracer instance)
         {
             try
             {
-                Assembly asm = Assembly.Load(new AssemblyName("SignalFx.Tracing.OpenTracing, Version=0.0.3.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb"));
+                Assembly asm = Assembly.Load(new AssemblyName("SignalFx.Tracing.OpenTracing, Version=0.1.1.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb"));
```
rmfitzpatrick, line 475: Looks like this replacement was dropped in adopting 0.1.0 changes: https://github.com/signalfx/signalfx-dotnet-tracing/blob/v0.0.3/tools/PrepareRelease/SetAllVersions.cs#L78 . Would you mind adding it back w/ these?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
```diff
@@ -172,7 +172,7 @@ public static class HttpMessageHandlerIntegration
             HttpRequestMessage request,
             CancellationToken cancellationToken)
         {
-            if (!(handler is HttpClientHandler || reportedType.FullName.Equals("System.Net.Http.SocketsHttpHandler", StringComparison.OrdinalIgnoreCase)) ||
+            if (!(handler is HttpClientHandler || "System.Net.Http.SocketsHttpHandler".Equals(reportedType.FullName, StringComparison.OrdinalIgnoreCase)) ||
```
rmfitzpatrick, line 175: Is this a style preference?

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -51,7 +51,13 @@ public static Scope CreateOutboundHttpScope(Tracer tracer, string httpMethod, Ur
                     return null;
                 }
 
-                scope = tracer.StartActive(OperationName);
+                var spanName = httpMethod ?? OperationName;
+                if (tracer.Settings.AppendUrlPathToName)
+                {
+                    spanName += ":" + requestUri.AbsolutePath;
```
rmfitzpatrick, line 57: Cardinality concerns would apply here right?  `/accounts/80d02dcc-2a0b-4f43-9f82-c5c501e9a797/update`

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
```diff
@@ -195,8 +195,15 @@ public static class HttpMessageHandlerIntegration
...
 
-                    // this tag can only be set after the response is returned
-                    scope?.Span.SetTag(Tags.HttpStatusCode, ((int)response.StatusCode).ToString());
+                    if (scope != null)
+                    {
+                        // this tag can only be set after the response is returned
+                        scope.Span.SetTag(Tags.HttpStatusCode, ((int)response.StatusCode).ToString());
+                        if (!response.IsSuccessStatusCode && !string.IsNullOrWhiteSpace(response.ReasonPhrase))
+                        {
+                            scope.Span.SetTag(Tags.HttpStatusText, response.ReasonPhrase);
```
rmfitzpatrick, line 204: Though unlikely, could custom reason phrases warrant truncation (e.g. error information)?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
pjanotti: Just because of `reportedType.FullName` in theory could be null so code analyzers complain about a potential null exception.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
pjanotti: It seems a good idea, just wondering if I should move that truncation to `SetTag`  itself.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
pjanotti: Yes, that's why it is optional and disabled by default. The thing is that going with HTTP method only is pretty useless (although very low cardinality). I should document somewhere this option and the implications. Any suggestion for the best place to document it?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
pjanotti: For now, will do it right here.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
pjanotti: @rmfitzpatrick @brianashby-sfx do we have a location documenting all the env var configurations? We could potentially look into deriving it directly from the code.

rmfitzpatrick: We should add a configuration detailing the env vars that aren't instrumentation-specific including service name, endpoint url, etc.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
rmfitzpatrick: when we add the configurable recorded value max length option it would be used in SetTag, but incidentally we've been using the truncate helper for its 1024 char hard limit.

PR #43, opened 2020-06-16
Adds AspNet Fx Instrumentations (Web API & MVC)

pjanotti:
Adds AspNet Fx Instrumentations for Web API & MVC.

Manually tested both on Windows.


src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -74,6 +74,11 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command, stri
         {
             span.Type = SpanTypes.Web;
             span.ResourceName = resourceName?.Trim();
+            if (!string.IsNullOrEmpty(span.ResourceName))
+            {
+                span.OperationName = span.ResourceName;
```
rmfitzpatrick, line 79: Did new test failures necessitate this?  This is generally done w/ zipkin encoding so there may have been fallout from the api type change.   Not against this, but preferred the controlled validation approach to better ensure cardinality is constrained.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -129,7 +143,12 @@ public static class AspNetWebApi2Integration
                         UpdateSpan(controllerContext, scope.Span);
                     }
 
-                    scope?.Span.SetException(ex);
+                    if (scope != null)
+                    {
+                        scope.Span.SetException(ex);
+                        scope.Span.SetTag(Tags.HttpStatusCode, "500");
```
rmfitzpatrick, line 149: Is there a way to guarantee this is the returned status code and not a best guess (and potentially a fabrication)?

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
pjanotti: No this was added for the new integrations. The code for the new integrations attempts to build the proper name for the resourceName and tries to avoid high-cardinality. Notice that the function where this is called is `DecorateWebServerSpan` which typically should have a proper route, so even if it is something like `/api/123456/discounts` (where the number is a user-id that is a parameter) it will come back as something like `/api/{userId}/discounts` for AspNet. That said the cleaning API removes numbers and GUIDs, however, some more convoluted combination where manual routing is done may hit some high-cardinality. Perhaps we have to separate the config options for client and server in this case: for client path disabled by default for server the reverse.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
pjanotti: Good question! I had forgotten about `HttpResponseException`! ASP.NET returns error 500 in case of exception except if an `HttpResponseException` is raised, in such cases it returns the code that was set on the exception. I will update the code to correctly handle it.

pjanotti: On second thought: it is possible for ASP.NET to change this behavior, the conservative approach is to only trust if it is an `HttpResponseException`.

pjanotti: `HttpResponseException` is not available on all versions, I will just remove this statement.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
pjanotti: I created an escape valve for this: another env var configuration. However, due to the way that the tests are organized it requires a refactor of the tests to make it testable. I will address testing this in a separate PR.

brianashby-sfx:
Should we update list of supported libraries?

pjanotti:
@brianashby-sfx we want to add 2 more integrations for .NET Framework and then we update it. It should happen during next sprint.

PR #51, opened 2020-07-15
Add sfx token support

pjanotti:
To work in scenarios like Lambda and Functions we need to support configuration of the sfx token. Adding this new env var we can then use two env vars to export directly to ingest: 

- SIGNALFX_ENDPOINT_URL
- SIGNALFX_ACCESS_TOKEN

@signalfx/instrumentation

### DIFF
--- a/src/Datadog.Trace/Agent/ZipkinContent.cs
+++ b/src/Datadog.Trace/Agent/ZipkinContent.cs
@@ -15,10 +15,14 @@ namespace Datadog.Trace.Agent
         private readonly ZipkinSerializer _serializer = new ZipkinSerializer();
         private readonly Span[][] _spans;
 
-        public ZipkinContent(Span[][] spans)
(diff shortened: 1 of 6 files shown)


src/Datadog.Trace/Agent/ZipkinContent.cs
```diff
@@ -15,10 +15,14 @@ internal class ZipkinContent : HttpContent
         private readonly ZipkinSerializer _serializer = new ZipkinSerializer();
         private readonly Span[][] _spans;
 
-        public ZipkinContent(Span[][] spans)
+        public ZipkinContent(Span[][] spans, string signalFxAccessToken)
```
rmfitzpatrick, line 18: is there a reason not to default to empty?

pjanotti: I wanted to make sure that if one is available it should be passed, so forcing the dev to be aware of it.

PR #53, opened 2020-07-23
Allow service name per span as an option

pjanotti:
In some scenarios, it is desirable to allow service name to be set per span and not globally by the process. This change adds an env var to allow that option.

Before this change "custom integration service names" were hidden because the span service name was ignored. Now we want these to follow the same as the process, otherwise, service name for these instrumentations will change if combined with manual instrumentation and the option to set service name per span.

@signalfx/instrumentation

### DIFF
--- a/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
+++ b/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
@@ -33,7 +33,7 @@ namespace Datadog.Trace.Configuration
         /// OpenTracing semantic convention.
         /// </summary>
         /// <seealso cref="TracerSettings.ServiceName"/>
-        public const string ServiceNamePerSpan = "SIGNALFX_SERVICE_NAME_PER_SPAN";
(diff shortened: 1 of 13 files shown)


src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -27,6 +27,14 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.ServiceName"/>
         public const string ServiceName = "SIGNALFX_SERVICE_NAME";
 
+        /// <summary>
+        /// Configuration key for the application's to allow spans to specify
+        /// different service names then the default one using the "service.name"
+        /// OpenTracing semantic convention.
+        /// </summary>
+        /// <seealso cref="TracerSettings.ServiceName"/>
+        public const string ServiceNamePerSpan = "SIGNALFX_SERVICE_NAME_PER_SPAN";
```
rmfitzpatrick, line 36: We may want to adopt the `_ENABLED` suffix for flags to cut back on ambiguity.

rmfitzpatrick:
Mind adding a readme update w/ example in custom instrumentation section?

PR #54, opened 2020-07-28
Enable WCF integration

pjanotti:
Enabling WCF instrumentation that already exists on DD. This is to just enable what already exists. There were no automated tests for this integration, this PR does not add one for it. All validations were made manually by enabling profiling for `Samples.Wcf` and `Samples.WcfClient` (they are built when running the cmd to test integrations).

The client-side for `WSHttpBinding` and `BasicHttpBinding` generates client spans via the `WebRequest` instrumentation. Enabling WCF integration creates a span for the server-side. The context is properly flowing between the client and the server. There is no client-side span for `NetTcpBinding`.

The server span looks like a generic HTTP server-side span in order to make clear that this is a WCF span the operation name will be pre-pended with `wcf.request`.

@signalfx/instrumentation

### DIFF
--- a/integrations.json
+++ b/integrations.json
@@ -1908,6 +1908,37 @@
     ]
   },
   {
+    "name": "Wcf",
+    "method_replacements": [
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.ServiceModel",
+          "type": "System.ServiceModel.Dispatcher.ChannelHandler",
+          "method": "HandleRequest",
+          "signature_types": [
+            "System.Boolean",
+            "System.ServiceModel.Channels.RequestContext",
+            "System.ServiceModel.OperationContext"
+          ],
+          "minimum_major": 4,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 4,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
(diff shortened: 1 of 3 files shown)


integrations.json
```diff
@@ -1907,6 +1907,37 @@
...
+    "name": "Wcf",
+    "method_replacements": [
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.ServiceModel",
+          "type": "System.ServiceModel.Dispatcher.ChannelHandler",
+          "method": "HandleRequest",
+          "signature_types": [
+            "System.Boolean",
```
johnbley, line 1920: I show HandleRequest not taking the boolean (only 2 params)?  From Framework 4.8.  Perhaps it's not a big deal, but also perhaps we need a guard/version lock?

pjanotti: The `bool` is the return type, see [here](https://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/Dispatcher/ChannelHandler.cs,940) and also:

https://github.com/signalfx/signalfx-dotnet-tracing/blob/4566d238ae1406585d86b0283185ebd1c80ffb93/src/Datadog.Trace.ClrProfiler.Managed.Core/InterceptMethodAttribute.cs#L73-82

johnbley: Make sense.

PR #58, opened 2020-08-06
Tracing library and version tags only on root spans

pjanotti:
Avoid redundant tags on all spans for the Tracing library name and version. Now only root spans will have that information.

@signalfx/instrumentation

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,5 @@
 <Project>
+  <!-- Modified by SignalFx -->
   <PropertyGroup>
     <IsPackable>false</IsPackable>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
--- a/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -108,12 +108,7 @@ namespace Datadog.Trace.Configuration
 
             Integrations = new IntegrationSettingsCollection(source);
 
-            GlobalTags = source?.GetDictionary(ConfigurationKeys.GlobalTags) ??
-                         // default value (empty)
-                         new ConcurrentDictionary<string, string>();
-
-            GlobalTags.Add(Tags.Language, TracerConstants.Language);
-            GlobalTags.Add(Tags.Version, TracerConstants.AssemblyVersion);
+            GlobalTags = source?.GetDictionary(ConfigurationKeys.GlobalTags);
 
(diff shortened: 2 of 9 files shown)


Directory.Build.props
```diff
@@ -1,4 +1,5 @@
 <Project>
+  <!-- Modified by SignalFx -->
```
pjanotti, line 2: Opportunistic fix

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -108,12 +108,7 @@ public TracerSettings(IConfigurationSource source)
 
             Integrations = new IntegrationSettingsCollection(source);
 
-            GlobalTags = source?.GetDictionary(ConfigurationKeys.GlobalTags) ??
-                         // default value (empty)
-                         new ConcurrentDictionary<string, string>();
-
-            GlobalTags.Add(Tags.Language, TracerConstants.Language);
-            GlobalTags.Add(Tags.Version, TracerConstants.AssemblyVersion);
+            GlobalTags = source?.GetDictionary(ConfigurationKeys.GlobalTags);
```
rmfitzpatrick, line 111: do we no longer want to fall back to an empty dictionary or is it never used?

pjanotti: `GlobalTags` can still be specified via the config, there is still a fallback empty dictionary but it now is closer to where it is needed: on the Zipkin serializer. In practice, this doesn't make a difference: it is still a single instance of a dictionary per process but it makes clear where a non-empty is actually needed, so if no `GlobalTags` were configured that one is `null`.

pjanotti: However, it did break some test that it didn't on my local run..., investigating...

test/Datadog.Trace.Tests/AgentWriterTests.cs
```diff
@@ -32,12 +33,12 @@ public async Task WriteTrace_2Traces_SendToApi()
             // TODO:bertrand it is too complicated to setup such a simple test
             var trace = new[] { new Span(_spanContext, start: null) };
             _agentWriter.WriteTrace(trace);
-            await Task.Delay(TimeSpan.FromSeconds(2));
+            await Task.Delay(TimeSpan.FromSeconds(3));
```
pjanotti, line 36: I suppose this test was not being run in my local procedures when I started to run it failed every time with the value 2 seconds, so changing it to 3.

PR #59, opened 2020-08-10
Add cfg to add client IP to ASP.NET server spans

pjanotti:
Adds a configuration setting to attempt to add "peer.ipv4" or "peer.ipv6" if the information is available for ASP.NET Core / Framework MVC and Web API 2.

By default it is disabled the env var "SIGNALFX_ADD_CLIENT_IP_TO_SERVER_SPANS" can be used to enable it.

This information is easily available on ASP.NET Core and since this is more in line with trends for .NET the case is "optimized" for it: legacy needs to create an IPAddress instance that is already available on ASP.NET Core. This simplifies the code while making it forward-looking.

@signalfx/instrumentation

### DIFF
--- a/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc4Tests.cs
+++ b/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc4Tests.cs
@@ -8,15 +8,17 @@ using Xunit.Abstractions;
 
 namespace Datadog.Trace.ClrProfiler.IntegrationTests
 {
-    public class AspNetMvc4Tests : TestHelper, IClassFixture<IisFixture>
+    public abstract class AspNetMvc4Tests : TestHelper
     {
         private readonly IisFixture _iisFixture;
+        private readonly bool _addClientIp;
 
-        public AspNetMvc4Tests(IisFixture iisFixture, ITestOutputHelper output)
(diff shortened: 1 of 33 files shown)


test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc4Tests.cs
```diff
@@ -8,15 +8,17 @@
 
 namespace Datadog.Trace.ClrProfiler.IntegrationTests
 {
-    public class AspNetMvc4Tests : TestHelper, IClassFixture<IisFixture>
+    public abstract class AspNetMvc4Tests : TestHelper
```
rmfitzpatrick, line 11: Is it simpler to use enumerable `MemberData` to avoid the separation requirement? (e.g. https://github.com/signalfx/signalfx-dotnet-tracing/blob/master/test/Datadog.Trace.ClrProfiler.IntegrationTests/MongoDbTests.cs#L17)

pjanotti: These tests start IIS and to make them less brittle and not to take so long waiting for IIS start/stop it is better to do all tests sharing the configuration in a single run of IIS. This won't scale if more configurations are added but then I think we should pursue a way to somehow dynamically change these settings. For an idea of the cold start of IIS, these tests almost always fail on the first run.

PR #61, opened 2020-08-11
Prepare Release 0.1.2.0

pjanotti:
Prepares release 0.1.2.0

@signalfx/instrumentation

### DIFF
--- a/.azure-pipelines/integration-tests.yml
+++ b/.azure-pipelines/integration-tests.yml
@@ -24,7 +24,7 @@ jobs:
         dotnetCoreSdkVersion: 3.0.x
         publishTargetFramework: netcoreapp3.0
       netcoreapp3_1:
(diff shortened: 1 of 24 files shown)


pjanotti:
For native_windows x64/x86 we hit a very fresh issue: https://developercommunity.visualstudio.com/content/problem/1145815/net-coremsbuild-tooling-not-finding-right-version.html

I don't think it is worth fixing it here: we need the CI image to be fixed or the SDK to properly install the dependencies.

pjanotti:
The integration-tests on Windows failed with an indication that a package restore is missing. I'm not sure why this didn't happen before, attempting trivial change to workaround it.

pjanotti:
In the end, I opted to change make AzurePipelines use a specific issue of the SDK to avoid the issue mentioned above.

PR #64, opened 2020-08-28
Reduce the probability of TraceId and SpanId collisions

pjanotti:
The `TraceId` and `SpanId` were 63-bit and generated from `Random` this had a relatively high chance of collision since the code created thread locals for the `Random` but let the system choose the seed, which can result in duplicate seeds due to the relative small range of seeds `(0, 2ˆ31 - 1)`.

- On .NET Framework: if the thread locals are created in rapid succession it is possible that multiple threads end up with the same seed. .NET Framework uses a low-resolution time seed (~15msec). Moreover, if the uptime of the system is over 49.8 days values can be repeated (not as relevant as the first issue).

- On .NET Core: the library uses a central random generator to seed the random object of each thread, however, there is no guarantee of seed uniqueness. It does require a number of threads to be created and destroyed to eventually hit a collision but it is possible. If some application creates and destroys threads at a fast pace it could hit this after some time. At around 54.5K threads creations, the chance of collision crosses 50%. A quick experiment confirmed it: the seed is typically repeated before 100K trials and typically much before that (the minimum in my trials was 1257 which is a relatively low number of threads to be created during the lifetime of an app)

- Possibly this is not an issue for original codebase because it typically operated with sampling reducing the chances of such collisions.

- The way to ensure the uniqueness of IDs on .NET is to use `Guid.NewGuid` is more expensive but avoiding collisions is more important. That said tests for various alternatives show that it is basically better than the current implementation, just losing to `Random` that isn't a viable solution due to the small range of seeds `(0, 2ˆ31 - 1)`.

Performance tests:

```terminal
Net48:
|             Method |      Mean |    Error |   StdDev |   StdErr |       Min |       Max |    Median |
|------------------- |----------:|---------:|---------:|---------:|----------:|----------:|----------:|
|             Random |  79.48 ns | 0.373 ns | 0.330 ns | 0.088 ns |  79.00 ns |  80.07 ns |  79.48 ns |
| NonContendedRandom |  93.18 ns | 1.044 ns | 0.977 ns | 0.252 ns |  91.81 ns |  95.06 ns |  92.87 ns |
|  ThreadLocalRandom | 110.68 ns | 2.204 ns | 2.061 ns | 0.532 ns | 108.66 ns | 114.36 ns | 109.48 ns |
|             RNGCsp | 214.45 ns | 3.188 ns | 2.982 ns | 0.770 ns | 211.71 ns | 220.52 ns | 213.17 ns |
|           FromGuid | 114.96 ns | 0.533 ns | 0.472 ns | 0.126 ns | 114.15 ns | 115.79 ns | 114.95 ns |
|       SafeFromGuid | 101.38 ns | 0.411 ns | 0.365 ns | 0.097 ns | 100.97 ns | 102.11 ns | 101.23 ns |


Net462:
|             Method |      Mean |    Error |   StdDev |   StdErr |       Min |       Max |    Median |
|------------------- |----------:|---------:|---------:|---------:|----------:|----------:|----------:|
|             Random |  80.75 ns | 0.403 ns | 0.337 ns | 0.093 ns |  80.23 ns |  81.31 ns |  80.70 ns |
| NonContendedRandom |  93.90 ns | 0.376 ns | 0.352 ns | 0.091 ns |  93.42 ns |  94.62 ns |  93.94 ns |
|  ThreadLocalRandom | 102.12 ns | 0.639 ns | 0.566 ns | 0.151 ns | 101.30 ns | 103.25 ns | 102.11 ns |
|             RNGCsp | 214.76 ns | 2.371 ns | 2.218 ns | 0.573 ns | 212.13 ns | 219.19 ns | 214.38 ns |
|           FromGuid | 116.68 ns | 0.697 ns | 0.618 ns | 0.165 ns | 115.94 ns | 118.22 ns | 116.52 ns |
|       SafeFromGuid | 101.76 ns | 0.760 ns | 0.674 ns | 0.180 ns | 100.67 ns | 103.03 ns | 101.53 ns |


Netcoreapp3.1:
|             Method |      Mean |    Error |   StdDev |   StdErr |       Min |       Max |    Median |
|------------------- |----------:|---------:|---------:|---------:|----------:|----------:|----------:|
|             Random |  92.77 ns | 0.602 ns | 0.533 ns | 0.143 ns |  91.98 ns |  93.76 ns |  92.64 ns |
| NonContendedRandom | 104.62 ns | 0.472 ns | 0.418 ns | 0.112 ns | 103.90 ns | 105.49 ns | 104.55 ns |
|  ThreadLocalRandom | 101.14 ns | 0.878 ns | 0.822 ns | 0.212 ns | 100.12 ns | 102.88 ns | 100.83 ns |
|             RNGCsp | 108.84 ns | 0.805 ns | 0.714 ns | 0.191 ns | 108.27 ns | 110.37 ns | 108.56 ns |
|           FromGuid |  94.76 ns | 0.710 ns | 0.629 ns | 0.168 ns |  94.00 ns |  95.90 ns |  94.60 ns |
|       SafeFromGuid |  93.04 ns | 0.315 ns | 0.263 ns | 0.073 ns |  92.66 ns |  93.40 ns |  93.00 ns |
````

Test code:
```c#
using System;
using System.Security.Cryptography;
using System.Threading;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

namespace SpanIDBenchmark
{
    [MinColumn, MaxColumn, MeanColumn, MedianColumn, StdDevColumn, StdErrorColumn]
    public class SpanIDGenerators
    {
        private readonly ThreadLocal<Random> _threadLocalRng = new ThreadLocal<Random>(() => new Random(42));
        private readonly Random _rng = new Random(42);
        private readonly object _rngLock = new Object();
        private readonly RNGCryptoServiceProvider _rngCsp = new RNGCryptoServiceProvider();
        private readonly byte[] _buffer = new byte[8];

        [Benchmark]
        public ulong Random()
        {
            var buffer = new byte[8];
            _rng.NextBytes(buffer);
            return BitConverter.ToUInt64(buffer, 0);
        }

        [Benchmark]
        public ulong NonContendedRandom()
        {
            var buffer = new byte[8];
            lock (_rngLock)
            {
                _rng.NextBytes(buffer);
            }
            return BitConverter.ToUInt64(buffer, 0);
        }

        [Benchmark]
        public ulong ThreadLocalRandom()
        {
            var buffer = new byte[8];
            _threadLocalRng.Value.NextBytes(buffer);
            return BitConverter.ToUInt64(buffer, 0);
        }

        [Benchmark]
        public ulong RNGCsp()
        {
            var buffer = new byte[8];
            _rngCsp.GetBytes(buffer);
            return BitConverter.ToUInt64(buffer, 0);
        }

        [Benchmark]
        public unsafe ulong FromGuid()
        {
            var buffer = new byte[8];
            var guid = Guid.NewGuid();
            var guidBytes = new ReadOnlySpan<byte>(&guid, sizeof(Guid));
            guidBytes.Slice(0, buffer.Length).CopyTo(buffer);

            return BitConverter.ToUInt64(buffer, 0);
        }

        [Benchmark]
        public ulong SafeFromGuid()
        {
            var guid = Guid.NewGuid();
            var guidBytes = guid.ToByteArray();

            return BitConverter.ToUInt64(guidBytes, 0);
        }
    }

    public class Program
    {
        public static void Main(string[] args)
        {
            var summary = BenchmarkRunner.Run(typeof(Program).Assembly);
        }
    }
}
```
```





@signalfx/instrumentation

### CODE REVIEW

File: src/Datadog.Trace/Sampling/RuleBasedSampler.cs [hunk 1]
```diff
@@ -82,7 +83,7 @@ public void RegisterRule(ISamplingRule rule)
 
         private SamplingPriority GetSamplingPriority(Span span, float rate)
         {
-            var sample = ((span.TraceId * KnuthFactor) % TracerConstants.MaxTraceId) <= (rate * TracerConstants.MaxTraceId);
+            var sample = ((span.SpanId * KnuthFactor) % TracerConstants.MaxTraceId) <= (rate * TracerConstants.MaxTraceId);
```
johnbley, line 86: Doesn't this change the semantics of sampling?  Before if a trace was sampled, all its spans were included.  This seems to break that?

rmfitzpatrick: I believe the calculation is essentially unchanged and this is only called when establishing the trace's sampling priority.

rmfitzpatrick:
Wholly adopting 128b trace ids will be a breaking change for propagation w/ libraries that don't support them (php and ruby).  What do you think about using last 64bits of a new guid and not moving from ulong?

pjanotti:
@rmfitzpatrick 
> Wholly adopting 128b trace ids will be a breaking change for propagation w/ libraries that don't support them (php and ruby). What do you think about using last 64bits of a new guid and not moving from ulong?

Thanks for bringing that up, I wasn't aware of that. An option to propagate only 64bits (off by default) will take care of it. Probably it will be the same id as the root span so it actually will be 2ˆ60 IDs - I will run the calculations but probably it is good enough.

src/Datadog.Trace/Sampling/RuleBasedSampler.cs
pjanotti: I wasn't concerned about this since we don't use this feature - I will take a second look at that to see how is it affected.

pjanotti:
Hum, actually may be better to have the default propagating 64bit IDs with an option to propagate the full 128bit: it needs to accept 128bit and 64bit but by default generating 64bit keeps things working between libraries without extra configurations.

rmfitzpatrick:
Are you saying that the .net trace id should be 128 but when propagated the remote services would see it as 64?  I think that would lead to broken traces in our service.

pjanotti:
> Are you saying that the .net trace id should be 128 but when propagated the remote services would see it as 64?

I wasn't clear: it should operate with 64-bit (effectively 60-bit) IDs by default (internally and as consequence for propagation) but there should be a way to use 128-bit trace IDs (we can receive them from other sources and should be able to handle them). That said it is possible to handle these things in separate PRs - still considering what path to take.

pjanotti:
Opted to keep using 64 bit IDs (in practice 60 bit), this will make the change much smaller and we can coordinate the approach of moving libraries to 128 bit support later.

src/Datadog.Trace/SpanContext.cs
```diff
@@ -96,5 +110,10 @@ private SpanContext(ulong? traceId, string serviceName)
         /// Gets or sets the span associated with this context.
         /// </summary>
         internal ISpan Span { get; set; }
+
+        private static ulong GenerateId()
+        {
+            return BitConverter.ToUInt64(Guid.NewGuid().ToByteArray(), 0) & RandomIdBitMask;
```
rmfitzpatrick, line 116: Would it be reasonable to get four additional random bits to fill in for those lost by the version?  The node section has 48 to pull from.

pjanotti: I had run some measurements of the perf impact of doing that and in some cases was almost 10% worst but bear in mind that it was in every case less than 10ns. The challenge was choosing one option that was "good" on all combinations of runtime and platform (but actually think that is reasonable to ignore x86). I will take a second look just in case.

johnbley:
👍 LGTM

rmfitzpatrick (approved):
good catch!

johnbley (approved):

PR #66, opened 2020-09-17
Add max "recorded" length

pjanotti:
Add max recorded length for tags and logs following the same pattern that was done for golang.

Moved the truncation to serialization assuming that most of the time nothing will be truncated and reducing the risk of allocating during actual operation being traced.

Separated truncation from SQL sanitation for "db.statement" tag.

### DIFF
--- a/src/Datadog.Trace/Agent/ZipkinSerializer.cs
+++ b/src/Datadog.Trace/Agent/ZipkinSerializer.cs
@@ -182,6 +182,29 @@ namespace Datadog.Trace.Agent
                     return annotations;
                 }
             }
+
+            // Methods below are used by Newtonsoft JSON serializer to decide if should serialize
+            // some properties when they are null.
+
+            public bool ShouldSerializeTags()
+            {	
+                return _tags != null;	
+            }	
+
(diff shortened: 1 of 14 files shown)


src/Datadog.Trace/Agent/ZipkinSerializer.cs
```diff
@@ -154,26 +182,6 @@ public string Kind
                     return annotations;
                 }
             }
-
-            public bool ShouldSerializeTags()
```
rmfitzpatrick, line 158: Are these safe to remove?  I vaguely recall them being needed to avoid null values in json, which I don't believe are allowed in zipkin v2.

pjanotti: Hum I didn't see any test complaining about it will take a second look to avoid surprises.

pjanotti: In principle we could remove the one for `Tags` and the one for `Annotations` but since we don't have a test directly for this I am going to put them back.

johnbley (approved):
LGTM

PR #73, opened 2020-10-13
Build an Azure Site Extension

pjanotti:
Builds an Azure Site Extension. Notice that at this stage the extension still needs to be published to nuget.org to automatically appear available to users on Azure - at this point, it can be added via a private NuGet feed (see https://github.com/projectkudu/kudu/wiki/Configurable-settings#site-extensions-gallery-feed-url).


pjanotti:
@rmfitzpatrick @johnbley @mateuszrzeszutek I just validated that the package generated via CI was working as expected, PTAL when you have a chance.

deploy/Azure.Site.Extension/README.md
```diff
@@ -0,0 +1,8 @@
+# SignalFx .NET Tracing Azure Site Extension
+
+Code on this folder is used to build a [Azure Site Extension](https://github.com/projectkudu/kudu/wiki/Azure-Site-Extensions)
```
rmfitzpatrick, line 3: nit, "in this folder"

tools/PrepareRelease/SetAllVersions.cs
```diff
@@ -114,6 +114,17 @@ public static void Run()
                 "deploy/Datadog.Trace.ClrProfiler.WindowsInstaller/Datadog.Trace.ClrProfiler.WindowsInstaller.wixproj",
                 WixProjReplace);
 
+            // Azure Site Extension updates
+            SynchronizeVersion(
+                "deploy/Azure.Site.Extension/applicationHost.xdt",
+                text => Regex.Replace(text, VersionPattern(), VersionString()));
+            SynchronizeVersion(
+                "deploy/Azure.Site.Extension/Azure.Site.Extension.nuspec",
+                text => Regex.Replace(text, VersionPattern(), VersionString()));
```
rmfitzpatrick, line 123: Should we be using `NuspecVersionReplace()`?

rmfitzpatrick (approved):
I have very little context for this but lgtm w/ items addressed.

tools/PrepareRelease/SetAllVersions.cs
pjanotti: Thanks, I didn't notice the existence of this method `NuspecVersionReplace`. Anyway, for this to work correctly it will need correct use of the `IsPreRelease` flag, this will grow the change quite a bit I will postpone when we automate more of the steps to cut a new release.

PR #74, opened 2020-10-26
Remove PCF tile

mateuszrzeszutek:
SignalFX tracing library won't have its own PCF tile, so I'm removing it from this repo -- it's both confusing (for the buildpack users) and unnecessary,

@signalfx/instrumentation


pjanotti (approved):
LGTM - sanity check: Linux stuff typically is not `zip` but it seems intentional and supported by `cfcreate-buildpack`.

mateuszrzeszutek:
> Linux stuff typically is not zip but it seems intentional and supported by cfcreate-buildpack.

Yep - CF uses zip as package format for buildpacks.

PR #75, opened 2020-11-18
ASP.NET Core MVC: handle incomplete route values

pjanotti:
I couldn't repro the exact situation on my machine but saw the logs from an Az App hitting that. Added a test to force the same case.

### DIFF
--- a/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
+++ b/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
@@ -168,7 +168,7 @@ namespace Datadog.Trace.DiagnosticListeners
                     remoteIp = httpContext?.Connection?.RemoteIpAddress;
                 }
 
-                span.DecorateWebServerSpan(null, httpMethod, host, url, remoteIp);
+                span.DecorateWebServerSpan(resourceUrl, httpMethod, host, url, remoteIp);
                 span.SetTag(Tags.InstrumentationName, IntegrationName);
 
                 // set analytics sample rate if enabled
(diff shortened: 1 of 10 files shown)


src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -203,19 +203,34 @@ private void OnMvcBeforeAction(object arg)
...
+                    string routeTemplate = actionDescriptor.AttributeRouteInfo?.Template;
+                    if (!string.IsNullOrEmpty(routeTemplate))
                     {
                         span.OperationName = routeTemplate;
                     }
+                    else if (!string.IsNullOrEmpty(controllerName) && !string.IsNullOrEmpty(actionName))
+                    {
+                        span.OperationName = $"{controllerName}.{actionName}".ToLowerInvariant();
+                    }
                 }
```
johnbley, line 234: The above sequence makes sense to me as a series of "use the best tag values available".

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/SpanTestHelpers.cs
```diff
@@ -45,9 +52,11 @@ public class SpanTestHelpers
                 }
             }
 
+            var spansMSg = Environment.NewLine + string.Join(Environment.NewLine, spanDump.Select(s => " * " + s));
```
johnbley, line 55: Tiny nit - `spansMsg` not `spasMSg` with its weird extra capital.

johnbley (approved):

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
pjanotti: Yes, that is correct. I will add a comment to make it clear.

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/SpanTestHelpers.cs
pjanotti: 🦅

PR #81, opened 2021-01-09
Rename public APIs

pjanotti:
Renaming the public APIs in order to have a single distribution for
manual and auto-instrumentation.


brianashby-sfx:
can we get an update to README.md to announce the change? i guess updating to the newer package will break implementations?

pjanotti:
> can we get an update to README.md to announce the change? i guess updating to the newer package will break implementations?

When we start to publish the Nuget package from CI of this repo I will add something to the README otherwise this change is completely transparent to the user.

dszmigielski:
Don't we want to rename the .csproj and the directory it sits in to SignalFx.Tracing as well to keep it consistent?

johnbley (changes requested):
I think this PR would benefit from another pass with grep; for example I found these in some out-of-the-way files that look like they need adjusting:
```
./reproductions/StackExchange.Redis.AssemblyConflict.LegacyProject/Program.cs:                var tracerAssemblyLocation = Type.GetType("Datadog.Trace.Tracer, SignalFx.Tracing")?.Assembly.Location ?? "(none)";
./reproductions/StackExchange.Redis.AssemblyConflict.SdkProject/Program.cs:                var tracerAssemblyLocation = Type.GetType("Datadog.Trace.Tracer, SignalFx.Tracing")?.Assembly.Location ?? "(none)";
./samples-aspnet/Samples.AspNetMvc5/Views/Home/Index.cshtml:                <td>@typeof(Datadog.Trace.Tracer).Assembly.Location</td>
./samples-aspnet/Samples.AspNetMvc4/Views/Home/Index.cshtml:                <td>@typeof(Datadog.Trace.Tracer).Assembly.Location</td>
./samples-aspnet/Samples.AspNetMvc5_0/Views/Home/Index.cshtml:                <td>@typeof(Datadog.Trace.Tracer).Assembly.Location</td>
./samples/Samples.AspNetCoreMvc21/Controllers/HomeController.cs:            ViewBag.TracerAssemblyLocation = Type.GetType("Datadog.Trace.Tracer, SignalFx.Tracing")?.Assembly.Location;
./samples/Samples.AspNetCore31/Startup.cs:                                                           var tracerAssemblyLocation = Type.GetType("Datadog.Trace.Tracer, Datadog.Trace")?.Assembly.Location;
./samples/Samples.AdditionalDiagnosticListeners/Controllers/HomeController.cs:            ViewBag.TracerAssemblyLocation = Type.GetType("Datadog.Trace.Tracer, Datadog.Trace")?.Assembly.Location;
```
I can help you with some command-line-fu if you'd like.

pjanotti:
@dszmigielski 
> Don't we want to rename the .csproj and the directory it sits into SignalFx.Tracing as well to keep it consistent?

The choice is to keep those in line with upstream but to be fair perhaps there is no/little advantage in doing that. Since it is not required I opted to start with them aligned with upstream. We can revisit it later if we think it will be better to rename.

@johnbley Thanks for the catch! I did it in chunks and clearly missed some locations especially the ones not tested by integration or manual tests.

PR #87, opened 2021-02-01
Elasticsearch7 support

dszmigielski:
Support for Elasticsearch7. Please note that we have a sort of regression here as the name for a Span was previously taken from one of the properties that is no longer available in version 7.x. I have also regenerated integrations.json and it seems some entries were removed, please verify if that is correct.


samples/Samples.Elasticsearch.V7/Program.cs
```diff
@@ -0,0 +1,428 @@
+﻿using System;
```
pjanotti, line 1: nit: I'm questioning the value of this practice but so far we have been adding the `// Modified by SignalFx` even to the files that we created :)

pjanotti (commented):
Looks good, I like the tests. The `integrations.json` needs to be corrected before merging.

PR #89, opened 2021-02-04
Add troubleshoot for custom instrumentation

pjanotti:
Adds a troubleshoot for a corner case between auto-instrumentation and custom-instrumentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ before accessing `GlobalTracer.Instance`.
         {
             try
             {
-                Assembly tracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing, Culture=neutral, PublicKeyToken=def86d061d0d2eeb"));
+                Assembly tracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing, Version=0.1.8.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb"));
                 Type tracerType = tracingAssembly.GetType("SignalFx.Tracing.Tracer");
 
                 PropertyInfo tracerInstanceProperty = tracerType.GetProperty("Instance", BindingFlags.Public | BindingFlags.Static);
--- a/README.md
+++ b/README.md
@@ -300,7 +300,8 @@ to include/exclude applications from the tracing auto-instrumentation.
 These are ";" delimited lists that control the inclusion/exclusion of processes.
 
 ### Custom instrumentation not being captured
-If the code accessing `GlobalTracer.Instance` is executed before any auto-instrumentation is injected
+
+If the code accessing `GlobalTracer.Instance` executes before anyg auto-instrumentation is injected
 into the process the call to `GlobalTracer.Instance` will return the OpenTracing No-Operation tracer.
 In this case it is necessary to force the injection of the SignalFx tracer by running a method like the one below
 before accessing `GlobalTracer.Instance`.
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ of your application, clients, and framework.
 
 1. Add the OpenTracing dependency to your project:
     ```xml
-    <PackageReference Include="OpenTracing" Version="0.12.1" />
+    <PackageReference Include="OpenTracing" Version="0.12.0" />
     ```
 2. Obtain the `OpenTracing.Util.GlobalTracer` instance and create spans that
 automatically become child spans of any existing spans in the same context:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -296,6 +296,31 @@ Use the environment variables `SIGNALFX_PROFILER_EXCLUDE_PROCESSES` and `SIGNALF
...
+```c#
+        static void InitTracer()
+        {
+            try
+            {
+                Assembly tracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing, Culture=neutral, PublicKeyToken=def86d061d0d2eeb"));
+                Type tracerType = tracingAssembly.GetType("SignalFx.Tracing.Tracer");
+
+                PropertyInfo tracerInstanceProperty = tracerType.GetProperty("Instance", BindingFlags.Public | BindingFlags.Static);
+                object tracerInstance = tracerInstanceProperty.GetValue(null);
```
pellared, line 314: I am thinking if it is worth to make some more precise error handling here.

1. We could catch `BadImageFormatException` which can be thrown by `Assembly.Load()`
2. `tracingAssembly.GetType("SignalFx.Tracing.Tracer")` can return null (almost impossible but I think if it would ever happen it would be better to have something else then `NullReferenceExpcetion`. `tracerType.GetProperty` can return and `tracerInstanceProperty.GetValue` can return nulls as well and it would no be normal.

pellared (commented):
Little nitpicking from my side. The stuff is probably not important, but I prefer to be "over-protective" in the beginning.

README.md
```diff
@@ -296,6 +296,31 @@ Use the environment variables `SIGNALFX_PROFILER_EXCLUDE_PROCESSES` and `SIGNALF
 to include/exclude applications from the tracing auto-instrumentation.
 These are ";" delimited lists that control the inclusion/exclusion of processes.
 
+### Custom instrumentation not being captured
+If the code accessing `GlobalTracer.Instance` is executed before any auto-instrumentation is injected
+into the process the call to `GlobalTracer.Instance` will return the OpenTracing No-Operation tracer.
+In this case it is necessary to force the injection of the SignalFx tracer by running a method like the one below
+before accessing `GlobalTracer.Instance`.
+
```
brianashby-sfx, line 304:
```suggestion
If the code accessing `GlobalTracer.Instance` executes before any
auto-instrumentation is injected into the process, the call to
`GlobalTracer.Instance` will return a OpenTracing no-operation tracer. If
the call returns the no-operation tracer, inject the SignalFx tracer by running
a method before accessing `GlobalTracer.Instance`. The following example
illustrates what this can look like:
```

README.md
```diff
@@ -225,7 +225,7 @@ of your application, clients, and framework.
 
 1. Add the OpenTracing dependency to your project:
     ```xml
-    <PackageReference Include="OpenTracing" Version="0.12.1" />
+    <PackageReference Include="OpenTracing" Version="0.12.0" />
```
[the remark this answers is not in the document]
pjanotti: The issue is that for legacy support the tracer DLLs are registered in the GAC bringing OpenTracing 0.12.0 so it doesn't work if the user app uses 0.12.1 - the original documentation was done against .NET Core that doesn't use the GAC and doesn't have this issue. 

I agree that it is worth documenting the info above but this section is intended for end-users and I want to keep it simple.

README.md
pjanotti: These are remote possibilities, except for `BadImageFormatException`, the way for the NRE to happen is if we change the type/method names. Handling those will add a bit of clutter to the code. In order to prevent the issue with certainty what we actually need is a test with the snippet to ensure that it works and then we link here directly to the snippet. I will add this to the backlog but at a lower priority.

pellared: TBH I am not sure if even handling `BadImageFormatException` is necessary. I remember that this exception can also happen when assemblies are references in a "normal" way (without exception).

Please treat my comment as extremely overprotective.

BTW what is the reason that the snippet uses reflection instead of usual reference and code invocation?

README.md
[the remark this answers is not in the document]
pellared: thx for explanation 👍

README.md
```diff
@@ -299,6 +299,32 @@ Use the environment variables `SIGNALFX_PROFILER_EXCLUDE_PROCESSES` and `SIGNALF
 to include/exclude applications from the tracing auto-instrumentation.
 These are ";" delimited lists that control the inclusion/exclusion of processes.
 
+### Custom instrumentation not being captured
+
+If the code accessing `GlobalTracer.Instance` executes before anyg auto-instrumentation is injected
```
pellared, line 304: typo 

```suggestion
If the code accessing `GlobalTracer.Instance` executes before any auto-instrumentation is injected
```

pellared (approved):

README.md
pjanotti: Let's chat later about this during this week.

README.md
pjanotti: good :eyes: fixed

PR #90, opened 2021-02-04
Automate release

dszmigielski:
Adding release automation to our CircleCI pipeline. Pushing tag should trigger the build that will add release assets and publish NuGet Packages.

To make it work we have to provide following environment variables to the CircleCI:
GITHUB_TOKEN - personal GitHub token; do we have a CI GitHub user?
NUGET_KEY - api key for accessing nuget.org
NUGET_FEED_URL - can change it to be static, if we will never use other feeds than nuget.org


.circleci/config.yml
```diff
@@ -50,23 +60,76 @@ jobs:
...
+      - PublishNuGetPackages:
+          requires:
+            - Alpine
+            - Linux
+            - Windows
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v\d+\.\d+\.\d+$/
```
pellared, line 134: I suggest removing the `$` at the end so that the automation would work e.g. for tags kuje `v1.2.3-rc` or `v.0.1.2-pre`

.circleci/config.yml
```diff
@@ -50,23 +60,76 @@ jobs:
...
   version: 2
   build:
     jobs:
-      - Alpine
-      - Linux
-      - Windows
+      - Alpine:
+          filters:
+            tags:
+              only: /.*/
```
pellared, line 106: add a comment that this is needed b/c otherwise this job would not be run when doing a release

.circleci/config.yml
```diff
@@ -50,23 +60,76 @@ jobs:
...
+      - attach_workspace:
+          at: ./
+      - run:
+          name: "Publish Release on GitHub"
+          command: |
+            ghr -t ${GITHUB_TOKEN} -u ${CIRCLE_PROJECT_USERNAME} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} ${CIRCLE_TAG} ./artifacts/
 
+  PublishNuGetPackages:
+    docker:
+      - image: mcr.microsoft.com/dotnet/sdk:3.1
```
pellared, line 90: any reason why not `sdk:5.0` ?

.circleci/config.yml
```diff
@@ -14,28 +14,38 @@ jobs:
...
       - run: docker-compose run build
       - run: docker-compose run Profiler
       - run: docker-compose run package
+      - run: mkdir artifacts
+      - run: cp -a deploy/linux/. artifacts/
       - store_artifacts:
-          path: deploy/linux/
+          path: artifacts/
+      - persist_to_workspace:
+          root: ./
```
pellared, line 41: `/` is not needed. See: https://circleci.com/docs/2.0/sample-config/. Applies to other places as well

.circleci/config.yml
```diff
@@ -50,23 +60,76 @@ jobs:
...
+          paths:
+              - artifacts
+
+  PublishRelease:
+    docker:
+      - image: cibuilds/github:0.10
+    steps:
+      - attach_workspace:
+          at: ./
```
pellared, line 82: `/` is not needed. See: https://circleci.com/docs/2.0/sample-config/. Applies to other places as well

.circleci/config.yml
```diff
@@ -50,23 +60,76 @@ jobs:
       - run: msbuild Datadog.Trace.proj /t:msi /p:Configuration=Release;Platform=x86
       - run: msbuild Datadog.Trace.proj /t:CreateHomeDirectory /p:Configuration=Release;Platform=x64
       - run: nuget pack -p NoWarn=NU5100;NoBuild=true;NoDefaultExcludes=true -OutputDirectory src\bin\Azure.Site.Extension deploy\Azure.Site.Extension\Azure.Site.Extension.nuspec
+      - run: mkdir artifacts
+      - run: cp -r C:\Users\circleci\project\deploy\Datadog.Trace.ClrProfiler.WindowsInstaller\bin\Release\x64\en-us\* artifacts
```
pellared, line 64: better replace `C:\Users\circleci\project\deploy` with `deploy` (as you it is done in Linux job)
one never knows when CIrcleCI would change the working path

pellared:
We might also like to add a manual confirmation before making a release: https://circleci.com/blog/manual-job-approval-and-scheduled-workflow-runs/

pellared:
Here I asked how we can get the needed tokens: https://splunk.slack.com/archives/CECPYFE0G/p1612449906005600

.circleci/config.yml
pjanotti: I like this idea but I would rather work it in a separate PR in case it requires changes to the prepare release tool or something else.

pjanotti (approved):

pjanotti:
A further idea for optimization will be to automate the creation of a PR to update the version after a label was created but let's avoid PR feature creep and consider it a separate change.

PR #106, opened 2021-06-09
Optimize default context propagation cases

pjanotti:
Optimizing the default propagation using B3 by focusing on "common" cases:

1. Inbound request but no B3 headers: in this case bailout earlier avoiding allocations and logging on every attempt to extract - the big saving, in this case, are due to the removal of one log per "extract"
2. Inbound request with B3 headers: saving on allocations by removing "early" allocations and enumerator allocations (not a big save in speed in this case, but some memory allocation)

Original:

|   Method |      Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |----------:|----------:|----------:|-------:|------:|------:|----------:|
|     NoB3 | 18.678 us | 0.3681 us | 0.9825 us | 0.4578 |     - |     - |   1.88 KB |
| B3Header |  1.193 us | 0.0235 us | 0.0344 us | 0.2708 |     - |     - |   1.11 KB |

Optimized:

|   Method |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
|     NoB3 |   419.0 ns |  8.23 ns | 15.47 ns | 0.0801 |     - |     - |     336 B |
| B3Header | 1,003.6 ns | 19.72 ns | 35.55 ns | 0.1926 |     - |     - |     808 B |

### DIFF
--- a/src/Datadog.Trace/Propagation/PropagationExtensions.cs
+++ b/src/Datadog.Trace/Propagation/PropagationExtensions.cs
@@ -17,24 +17,6 @@ namespace SignalFx.Tracing.Propagation
             return propagator.Extract(headers, ExtractFromHeadersCollection);
         }
 
-        public static IEnumerable<KeyValuePair<string, string>> ExtractHeaderTags(this IHeadersCollection headers, IEnumerable<KeyValuePair<string, string>> headerToTagMap)
-        {
-            foreach (KeyValuePair<string, string> headerNameToTagName in headerToTagMap)
-            {
-                var headerValue = ParseString(headers, headerNameToTagName.Key);
-
-                if (headerValue != null)
-                {
-                    yield return new KeyValuePair<string, string>(headerNameToTagName.Value, headerValue);
-                }
-            }
-        }
-
-        public static string ParseString(this IHeadersCollection headers, string headerName)
-        {
-            return PropagationHelpers.ParseString(headers, (carrier, header) => carrier.GetValues(header), headerName);
-        }
-
         private static void InjectToHeadersCollection(IHeadersCollection carrier, string header, string value)
         {
             carrier.Set(header, value);
--- a/src/Datadog.Trace/Propagation/PropagationHelpers.cs
+++ b/src/Datadog.Trace/Propagation/PropagationHelpers.cs
@@ -1,5 +1,4 @@
 // Modified by SignalFx
-
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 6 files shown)


pellared:
What have you used for benchmarking. https://github.com/dotnet/BenchmarkDotNet ?

src/Datadog.Trace/Propagation/PropagationExtensions.cs
```diff
@@ -17,24 +17,6 @@ public static SpanContext Extract(this IPropagator propagator, IHeadersCollectio
             return propagator.Extract(headers, ExtractFromHeadersCollection);
         }
 
-        public static IEnumerable<KeyValuePair<string, string>> ExtractHeaderTags(this IHeadersCollection headers, IEnumerable<KeyValuePair<string, string>> headerToTagMap)
```
pellared, line 20: it was unused?

src/Datadog.Trace/Propagation/PropagationHelpers.cs
```diff
@@ -10,30 +9,29 @@ internal static class PropagationHelpers
...
             }
 
-            logger.Warning("Could not parse {HeaderName} headers: {HeaderValues}", headerName, string.Join(",", headerValues));
-            return TraceId.Zero;
-        }
-
-        public static string ParseString<T>(T carrier, Func<T, string, IEnumerable<string>> getter, string headerName)
-        {
-            var headerValues = getter(carrier, headerName);
+            var headerValues = enumerableHeaderValues.ToList();
```
pellared, line 18: nit: https://stackoverflow.com/a/60103725
probably not worth to change but just for knowledge sharing 😉

pellared (approved):

src/Datadog.Trace/Propagation/PropagationExtensions.cs
pjanotti: Correct.

pjanotti:
> What have you used for benchmarking. https://github.com/dotnet/BenchmarkDotNet ?

Exactly.

src/Datadog.Trace/Propagation/PropagationHelpers.cs
pjanotti: I'm glad that you mentioned this: my knowledge is out of date in this regard: I had some vague memory that if the underlying object was a List<T> the ToList() was optimized but measuring against ToArray the latter save 32 bytes and keeps the same speed :) so changed it.

pjanotti: to be clear: 32 bytes per call not for the whole benchmark,

PR #111, opened 2021-06-15
Improve IIS hosted apps cardinality

RassK:
### Background

Regarding to SWAT: https://signalfuse.atlassian.net/browse/SWAT-3239

Porting some code from OTel solution to improve IIS hosted apps cardinality.
Resource name will be replaced with route template when optional feature `SIGNALFX_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` is `true`.

### Additionally
- Upgrades Linux from 16.04 => 18.04
- Upgrades .NET SDK to 5.0.x


src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -337,21 +374,9 @@ public static Scope CreateScope(object controllerContext)
                     return res;
                 }
 
-                var response = httpContext?.Response;
```
pjanotti, line 340: Why not keep the local and avoid the two calls if it is null? It seems more readable than using the null-conditional operator in both calls.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -97,21 +101,52 @@ public static Scope CreateScope(object controllerContext)
...
-                string actionName = (routeValues?.GetValueOrDefault("action") as string)?.ToLowerInvariant();
-
                 if (string.IsNullOrEmpty(resourceName))
                 {
                     // Keep the legacy resource name, just to have something
                     resourceName = $"{httpMethod} {controllerName}.{actionName}";
                 }
 
+                // Replace well-known routing tokens
+                resourceName =
```
pjanotti, line 127: Perhaps worth measuring with BenchmarkDotNet if it would be worth to have `resourceName` as a `StringBuilder` due to the multiple `Replace` calls potentially allocating and recopying the string. This should be done with the typical case in mind: i.e., if the typical case nothing is replaced probably it is better to keep as it is.

Notice that this should be only called if resourceName was not null or empty prior to the checks above (starting on line 109) so it seems worth a single check for that at line 109.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -200,64 +201,73 @@ private static void UpdateSpan(dynamic controllerContext, Span span)
         {
             try
             {
-                var req = controllerContext?.Request as HttpRequestMessage;
+                var newResourceNamesEnabled = Tracer.Instance.Settings.RouteTemplateResourceNamesEnabled;
```
pjanotti, line 204: nit: declare and initialize near the first usage.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -200,64 +201,73 @@ private static void UpdateSpan(dynamic controllerContext, Span span)
...
                 }
                 catch
                 {
                 }
 
-                // Fail safe to catch templates in routing values
+                // Replace well-known routing tokens
                 resourceName =
                     resourceName
+                       .Replace("{area}", area)
```
pjanotti, line 256: Similar comment about possible use of `StringBuilder` and also avoiding calling the Replace methods if `resourceName` doesn't contain any of the tokens.

test/Directory.Build.props
```diff
@@ -4,6 +4,9 @@
   <PropertyGroup>
     <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
+
+    <!-- Hide warnings for EOL .NET Core targets (e.g. netcoreapp3.0) -->
```
pjanotti, line 8: Good! Don't we have the same warnings for code under `src/`? If yes perhaps move it one level up.

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -138,6 +138,14 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
                 scope.Span.DecorateWebServerSpan(resourceName, httpMethod, host, url, remoteIp);
 
                 httpContext.Items[_httpContextScopeKey] = scope;
+
+                // Decorate the incoming HTTP Request with distributed tracing headers
+                // in case the next processor cannot access the stored Scope
+                // (e.g. WCF being hosted in IIS)
+                if (HttpRuntime.UsingIntegratedPipeline)
```
pjanotti, line 145: Not sure if I understand which processor is going to retrieve this header?

pjanotti (commented):
Nice. Good that you updated the c# compiler version.

Not from your changes but the code to handle the resource name seems overly complex with a lot of fallbacks and conditions. Perhaps later we should try to simplify it later.

Directory.Build.props
```diff
@@ -4,7 +4,7 @@
     <IsPackable>false</IsPackable>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
-    <LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
RassK: https://docs.microsoft.com/en-gb/dotnet/csharp/language-reference/configure-language-version#defaults
Visual studio version perhaps they switched to explicit version in OTel?

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -160,6 +168,19 @@ private void OnEndRequest(object sender, EventArgs eventArgs)
...
+
+                    if (app.Context.Items[SharedConstants.HttpContextPropagatedResourceNameKey] is string resourceName
+                        && !string.IsNullOrEmpty(resourceName))
+                    {
+                        scope.Span.ResourceName = resourceName;
+                    }
+                    else
+                    {
+                        string path = UriHelpers.GetCleanUriPath(app.Request.Url);
+                        scope.Span.ResourceName = $"{app.Request.HttpMethod.ToUpperInvariant()} {path.ToLowerInvariant()}";
```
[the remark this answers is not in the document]
RassK: Currently it seems like a common pattern to lowercase paths. If APM portal can handle casing then could be simplified, otherwise if it couldn't then it's important for cardinality.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -337,21 +374,9 @@ public static Scope CreateScope(object controllerContext)
...
-                }
-                else if (response?.StatusCode != null)
-                {
-                    scope.Span.SetTag(Tags.HttpStatusCode, response.StatusCode.ToString(CultureInfo.InvariantCulture));
-                    if (!string.IsNullOrWhiteSpace(response?.StatusDescription))
-                    {
-                        scope.Span.SetTag(Tags.HttpStatusText, response.StatusDescription);
-                    }
-                }
+                scope.Span.SetHttpStatusCode(httpContext?.Response);
```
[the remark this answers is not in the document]
RassK: these integrations are for FX only. Notice the `#if !NETSTANDARD2_0` at the beginning of every integration.
+ It should have failed build if it was wrong way.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -91,6 +93,24 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command, stri
             }
         }
 
+#if NETFRAMEWORK
```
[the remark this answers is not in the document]
RassK: Net standard does not have HttpResponse (from system.web).

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -91,6 +93,24 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command, stri
...
+        internal static void SetHttpStatusCode(this Span span, HttpResponse response)
+        {
+            if (response?.StatusCode == 200)
```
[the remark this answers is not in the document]
pellared: BTW could it ever be `null`?

src/Datadog.Trace.AspNet/TracingHttpModule.cs
RassK: I guess it's not the next processor in pipeline but some bytecode instrumentation processor.

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -24,40 +29,113 @@ public static string CleanUri(Uri uri, bool removeScheme, bool tryRemoveIds)
             return $"{uri.Scheme}{Uri.SchemeDelimiter}{uri.Authority}{path}";
         }
 
+        [Obsolete("This method is deprecated and will be removed. Use GetCleanUriPath() instead. " +
```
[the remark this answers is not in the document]
RassK: Removed.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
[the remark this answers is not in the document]
RassK: Tried a bit more modern way, plus added some more commonly used codes.

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -24,40 +29,113 @@ public static string CleanUri(Uri uri, bool removeScheme, bool tryRemoveIds)
...
+        public static string GetCleanUriPath(string absolutePath)
         {
-            if (string.IsNullOrWhiteSpace(segment))
+            if (string.IsNullOrWhiteSpace(absolutePath) || (absolutePath.Length == 1 && absolutePath[0] == '/'))
+            {
+                return absolutePath;
+            }
+
+            // Sanitized url will be at worse as long as the original
+            var sb = StringBuilderCache.Acquire(absolutePath.Length);
```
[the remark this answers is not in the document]
RassK: Not sure I follow what do you mean? - feel free to push your idea as a commit.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
RassK: Pushed an example for SB usage, did you mean something like that?

README.md
```diff
@@ -73,6 +73,7 @@ Use these environment variables to configure the tracing library:
 | `SIGNALFX_STDOUT_LOG_ENABLED` | `false` | Enables `stdout` logging. This is disabled by default. |
 | `SIGNALFX_SYNC_SEND` | `false` | Enable to send spans in synchronous mode when the root span is closed. Sending spans in synchronous mode is generally recommended for only tests, but can also be useful for some special scenarios.|
 | `SIGNALFX_TRACE_DOMAIN_NEUTRAL_INSTRUMENTATION` | `false` |  Sets whether to intercept method calls when the caller method is inside a domain-neutral assembly. This is recommended when instrumenting IIS applications. |
+| `SIGNALFX_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` | `false` |  Feature Flag: enables updated resource names on `aspnet.request`, `aspnet-mvc.request`, `aspnet-webapi.request`, and `aspnet_core.request` spans. Enables `aspnet_core_mvc.request` spans and additional features on `aspnet_core.request` spans. |
```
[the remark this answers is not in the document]
RassK: I'm not sure how much we should diverge from OTel `OTEL_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED`.
It might be long but less confusion when working with SFx and OTel.

Directory.Build.props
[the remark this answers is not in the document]
pjanotti: Thanks for the link. I think the issue is with VS: "However, once you upgrade to Visual Studio 2019, that means the latest released version, regardless of the target framework.". The regardless part seems to be the problem. Let's go with 9.0

README.md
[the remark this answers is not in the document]
pjanotti: OTel was not released yet and the name is quite long, although many times will copy it from the docs. That said a shorter name if possible is better. Where that name is used on OTel?

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
pjanotti: You could have simply a thread static with the sb and the effect will be the same without the need for a new type for the cache. Let's have a quick chat about it to clarify.

src/Datadog.Trace.AspNet/SharedConstants.cs
```diff
@@ -0,0 +1,7 @@
+namespace Datadog.Trace.AspNet
+{
+    internal static class SharedConstants
+    {
+        public const string HttpContextPropagatedResourceNameKey = "__SignalFx.Tracing.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename";
```
pjanotti, line 5: I didn't find this being used anywhere in the repo? Am I missing something?

src/Datadog.Trace.AspNet/TracingHttpModule.cs
pjanotti: It feels strange to an inbound to inject but assuming that some other inbound may be under this one it makes sense.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -97,21 +101,53 @@ public static Scope CreateScope(object controllerContext)
...
-
                 if (string.IsNullOrEmpty(resourceName))
                 {
                     // Keep the legacy resource name, just to have something
                     resourceName = $"{httpMethod} {controllerName}.{actionName}";
                 }
 
+                // Replace well-known routing tokens
+                var resourceNameBuilder = new StringBuilder(resourceName);
+                resourceNameBuilder
```
pjanotti, line 129: nit: remove this line and the `;` from the previous should look cleaner.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -97,21 +101,53 @@ public static Scope CreateScope(object controllerContext)
...
+                       .Replace("{area}", areaName)
```
pjanotti, line 130: The 3 if statements above have in their condition `string.IsNullOrEmpty(resourceName)` so it seems that all these replace calls are only worth if none of the if statement conditions were satisfied, otherwise these replaces are going be called and we already know that no match will be found.

src/Datadog.Trace.AspNet/SharedConstants.cs
RassK: Did you copy the name with space at the end by accident? 
Otherwise there are references in `TracingHttpModule.cs`, `AspNetMvcIntegration.cs`, `AspNetWebApi2Integration.cs`

README.md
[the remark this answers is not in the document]
RassK: The same section: `ConfigurationKeys.FeatureFlags`. We should port the modifications to OTel also then.
I'll replace with advised name. 

The other important thing I forgot. In OTel it's by default false, but should we change it to true by default? Because APM portal would expect this by default working?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
RassK: Rearranged these blocks... seems that it's needed for only one type of routing.

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
RassK: But the new type also takes care of removing SB from cache if it's aquired for thread. In case there are 2 aquires for the same thread, it will mess up the SB instance.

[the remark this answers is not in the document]
pjanotti: It can't have two acquires the same thread since there is no async switch when the SB is used.

src/Datadog.Trace.AspNet/SharedConstants.cs
pjanotti: Yeah, somehow I missed it.

README.md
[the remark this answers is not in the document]
pjanotti: We should do both things on OTel.: 1. because it is a better name; 2. Notice that default right now it is actually false: https://github.com/signalfx/signalfx-dotnet-tracing/pull/111/files#diff-5c417e25757186c840db429ba030b837d74ca21b2fcfdb9a06f9b96cfc24187eR177

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
pjanotti: I was thinking in doing a quick validation with [BenchmarkDotNet](https://benchmarkdotnet.org/) about perf on this case.

pjanotti: It is ok to skip it for this PR.

pjanotti:
It seems that the code selecting ports already starts very high and then only goes up not and doesn't re-use ports used in previous tests. Re-kicking failed https://dev.azure.com/signalfx/767e0a0b-7429-47d0-b68e-26770411d549/_apis/build/builds/5499/logs/88

pjanotti (approved):

pjanotti:
Same again https://dev.azure.com/signalfx/767e0a0b-7429-47d0-b68e-26770411d549/_apis/build/builds/5499/logs/107 - merging and changing it later.

PR #113, opened 2021-06-17
2 tiny optimizations: most spans have tags and avoid allocating Stopwatch

pjanotti:
1. Most spans have Tags but not Logs: optimize for this case allocate a Dictionary with capacity that avoids re-sizing for most of the cases and only allocate Logs if it is actually used.
2. Use the timestamp directly instead of the the Stopwatch since the later is a reference type.

### DIFF
--- a/src/Datadog.Trace/Agent/ZipkinSerializer.cs
+++ b/src/Datadog.Trace/Agent/ZipkinSerializer.cs
@@ -155,7 +155,7 @@ namespace SignalFx.Tracing.Agent
             {
                 get
                 {
-                    if (_span.Logs.Count == 0)
+                    if (!_span.HasLogs || _span.Logs.Count == 0)
                     {
                         return emptyAnnotations;
                     }
--- a/src/Datadog.Trace/Agent/ZipkinSerializer.cs
+++ b/src/Datadog.Trace/Agent/ZipkinSerializer.cs
@@ -21,7 +21,7 @@ namespace SignalFx.Tracing.Agent
 
         public static IDictionary<string, string> BuildTags(Span span, TracerSettings settings)
         {
-            var tags = new Dictionary<string, string>();
+            var tags = new Dictionary<string, string>(span.Tags.Count);
             var recordedValueMaxLength = settings.RecordedValueMaxLength;
             foreach (var entry in span.Tags)
             {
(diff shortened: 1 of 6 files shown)


src/Datadog.Trace/Agent/ZipkinSerializer.cs
```diff
@@ -21,7 +21,7 @@ internal class ZipkinSerializer
 
         public static IDictionary<string, string> BuildTags(Span span, TracerSettings settings)
         {
-            var tags = new Dictionary<string, string>();
+            var tags = new Dictionary<string, string>(span.Tags.Count);
```
pjanotti, line 24: The serialization could avoid a bunch of allocations by using a custom serializer. Perhaps this will be next in the list.

RassK (approved):

src/Datadog.Trace/Span.cs
```diff
@@ -87,11 +92,13 @@ public string ServiceName
 
         internal TimeSpan Duration { get; private set; }
 
-        internal Dictionary<string, string> Tags { get; private set; }
+        internal IDictionary<string, string> Tags { get; } = new Dictionary<string, string>(16);
```
pellared, line 95: nit:
any reason why `IDictionary` not just `Dictionary`? 
I remember that accessing via the interface is a little slower 😄

pjanotti: Short answer: left over of an experiment that I was doing :) thanks for noticing it.

Long answer: instead of a dictionary I had a type that internally used an array until a certain size (I played a lot with 16) at which point it switched to a dictionary. It does optimize memory and it is a bit faster for the "typical" case but I'm for now I think I will shelve that idea (the gain is small and it likely deservers a tuning configuration, this last bit made me give up of it for the time being).

pjanotti: That said notice that this is exposed as a Dictionary for the benefit of serialization, the rest of the code most should be using SetTag/GetTag.

src/Datadog.Trace/Span.cs
```diff
@@ -33,11 +35,14 @@ internal Span(SpanContext context, DateTimeOffset? start)
             ServiceName = context.ServiceName;
             StartTime = start ?? Context.TraceContext.UtcNow;
 
-            Logger.Debug(
-                "Span started: [s_id: {0}, p_id: {1}, t_id: {2}]",
-                SpanId,
-                Context.ParentId,
-                TraceId);
+            if (IsLogLevelDebugEnabled)
```
RassK, line 38: What's the reason that IsLogLevelDebugEnabled is brought to `Span` class explicitly while all Loggers check it anyway?

pellared (approved):

PR #115, opened 2021-06-17
Performance pitfall in W3C propagation

dszmigielski:
*What*
Removing unnecessary logging and changing level from WARNING to DEBUG to improve performance.

*Why*
When enabling W3C propagation server side will hit performance pitfall due to excessive logging if the header is missing.


src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -51,7 +51,6 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
             var traceParentCollection = getter(carrier, W3CHeaderNames.TraceParent).ToList();
```
pjanotti, line 51: This should follow the same pattern at https://github.com/signalfx/signalfx-dotnet-tracing/blob/bff4458480463a75d870b41504eec6f08294ed47/src/Datadog.Trace/Propagation/PropagationHelpers.cs#L10-L35

See PR https://github.com/signalfx/signalfx-dotnet-tracing/pull/106

pjanotti: This avoids allocating for a very common case: no incoming headers.

src/Datadog.Trace/TraceId.cs
```diff
@@ -114,6 +116,7 @@ public static TraceId CreateFromString(string id)
             }
             catch (Exception ex) when (ex is ArgumentOutOfRangeException || ex is InvalidOperationException || ex is OverflowException || ex is FormatException)
             {
+                Log.Debug($"Could not parse TraceId from string {id}. {ex.Message}");
```
pjanotti, line 119: Adding this here will double the logging for B3 propagation in case of error. It seems that we need two APIs: `TryParse` and `Parse` instead of `CreateFromString` following the traditional pattern for .NET.  As it is it logs error if the string had length 16 or 32 and failed to parse but not if it had an invalid length.

In general this type of logging should be on the higher level not on a type that is a "primitive" type.

pjanotti (changes requested):
We can still avoid a few allocations on the typical case (no incoming headers). The API suggestion for TraceId is not required but logging seems better if out of it: we can give better contextual information about the type of header.

src/Datadog.Trace/TraceId.cs
dszmigielski: I will change that tomorrow, as changing the API requires a bit more caution so I don't break the rest of the code.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
dszmigielski: Done

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -48,26 +48,24 @@ public void Inject<T>(SpanContext context, T carrier, Action<T, string, string>
...
-            if (traceParentCollection.Count != 1)
+            var enumerableHeaderValues = getter(carrier, W3CHeaderNames.TraceParent);
+            if (enumerableHeaderValues == Enumerable.Empty<string>())
             {
-                Log.Warning("Header {HeaderName} needs exactly 1 value", W3CHeaderNames.TraceParent);
                 return null;
             }
 
-            var traceParentHeader = traceParentCollection.First();
+            var traceParentHeader = enumerableHeaderValues.First();
```
pjanotti, line 57: This makes me wonder if we can assume single header value for B3 too... we should look into that, and perhaps make it configurable but this is out of scope for this PR.

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
```diff
@@ -48,26 +48,29 @@ public void Inject<T>(SpanContext context, T carrier, Action<T, string, string>
...
             var traceIdString = traceParentHeader.Substring(VersionPrefixIdLength, TraceIdLength);
-            var traceId = TraceId.CreateFromString(traceIdString);
+            if (!TraceId.TryParse(traceIdString, out var traceId))
```
RassK, line 59: just a nit: I typically do not declare the out var inline when it's not used in the declared scope. Although technically it's not scope leaking, for readability imo it's for sure scope leaking. What others think about that feature? @pjanotti @pellared

pellared: personally, I do not like `out var` at all 😄

dszmigielski: I like it tbh 😄 I can see why you can find it unreadable though.

pjanotti: I don't have anything against `out var` sometimes it can be benefit of an explicit type but in the case of a `TryParse` it is clear because of the type must be before the method. Scope usage is kind of related I don't have strong opinion in this case so if people prefer to declare outside it is fine for me.

pjanotti (approved):
LGTM

src/Datadog.Trace/Propagation/W3CSpanContextPropagator.cs
pellared: I do not have a strong opinion as well

src/Datadog.Trace/TraceId.cs
```diff
@@ -83,38 +84,51 @@ public static TraceId CreateRandom()
...
-                        var lower = Convert.ToUInt64(lowerAsString, fromBase: 16);
-
-                        return new TraceId(higher, lower);
-                    }
-
-                    default:
-                    {
-                        return Zero;
-                    }
+                    throw new FormatException("The passed string must have 16 or 32 signs");
```
pellared, line 110: I think "character" is a better word than "sign". In Polish it is the same word but I think that in English sign is a little different

src/Datadog.Trace/TraceId.cs
```diff
@@ -83,38 +84,51 @@ public static TraceId CreateRandom()
         /// </summary>
         /// <param name="id">16 or 32 sign string ID to be parsed.</param>
         /// <returns>Instance of <see cref="TraceId"/> representing the same traceId as the passed string.</returns>
-        public static TraceId CreateFromString(string id)
+        public static TraceId Parse(string id)
```
pellared, line 87: Cannot we return `TraceId.Zero` instead of throwing the exception?
I think this method is not used anywhere except tests and `TryParse` and most often if `TryParse` returns `false` than there is a fallback to set `TraceId.Zerp` instead (which is not needed BTW).

We can sync if I am not clear

dszmigielski: Logging was an issue. I didn't know if someone passed `TraceId` equal to `0` or something that failed to parse - we log these differently. The fallback was just a safeguard for me if I miss some usages (so the behaviour stays the same).

pellared (approved):

pjanotti (approved):

PR #120, opened 2021-06-22
WIP: *DO NOT MERGE* Serialization optimizations

pjanotti:
Optimizes serialization of spans. This is not expected to reduce latency of single operations but the overall impact of the tracer on the application by optimizing the serialization of spans: it is faster and uses less memory, however, the memory lives longer (typically reaching gen2). Measurements under load indicate that overall results are better.


pjanotti:
These are good optimizations but not priority at this moment. We can revisit if needed.

PR #126, opened 2021-06-25
Instrument SqlCommand ExecuteXmlReader and ExecuteXmlReaderAsync

pjanotti:
Add instrumentation to two SqlCommand methods: ExecuteXmlReader and ExecuteXmlReaderAsync. This was a gap for SqlCommand in which all other command execution paths were covered except for these 2.


dszmigielski (approved):
LGTM

PR #127, opened 2021-06-29
Add basic Produce and Consume instrumentation

dszmigielski:


src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/KafkaHeadersCollectionAdapter.cs
```diff
@@ -0,0 +1,67 @@
...
+        private readonly MethodInfo _add;
+        private readonly MethodInfo _remove;
+
+        public KafkaHeadersCollectionAdapter(object headers)
+        {
+            _headers = headers;
+            var methods = _headers.GetType().GetMethods().ToList();
+            _tryGetLastBytes = methods.FirstOrDefault(m => m.Name == "TryGetLastBytes");
+            _add = methods.FirstOrDefault(m => m.Name == "Add" && m.GetParameters().Length == 2);
+            _remove = methods.FirstOrDefault(m => m.Name == "Remove");
```
RassK, line 26: `FirstOrDefault` can generate unhandled nullreference exceptions due caller doesn't verify if `MethodInfo` exists. 
I see 2 options:
1. use null check with methodinfo.invoke
2. use `First()` instead of `FirstOrDeafult()` and fail integration.  (I would prefer this, so it clearly indicates that something is wrong with integration).

dszmigielski: I would rather not fail the whole thing. Rather just log that we can't get headers. We can proceed with span creation it just won't have propagated context.

dszmigielski: I have added exception handling. PTAL

RassK: @pellared once suggested not to use generalized exception handling... but to target more specifically concrete failing point. 

+ I think logger with exception overload should be used, not to miss exception info.

RassK: try / catch instead of null check would add massive performance degradation. Usually these systems (having messaging middleware) are under heavy load - so it would have a big effect. 

Taking account it would happen also when method signature changes. Maybe an off switch with warning is more appropriate.

src/Datadog.Trace/Tags.cs
```diff
@@ -221,5 +221,94 @@ public static class Tags
...
+            /// </remarks>
+            public const string Tombstone = "messaging.kafka.tombstone";
+
+            /// <summary>
+            /// The Kafka client name.
+            /// </summary>
+            /// <remarks>
+            /// OpenTelemetry experimental specification at commit 5a19b53d71e967659517c02a69b801381d29bf1e.
+            /// </remarks>
+            public const string ClientName = "massaging.kafka.client_id";
```
dszmigielski, line 271: For this we are using "Name" property, which according to documentation is not a client_id:
```
Gets the name of this client instance.
Contains (but is not equal to) the client.id
configuration parameter.
```

samples/Samples.Kafka/Program.cs
```diff
@@ -1,47 +1,209 @@
...
+                });
+
+            using (var consumer = consumerBuilder.Build())
             {
-                consumer.Subscribe(topic);
+                consumer.Subscribe(topicName);
+
+                // The Kafka calls are all actually happen in the background give it a delay
+                // to ensure that the first consume call works.
+                Thread.Sleep(DefaultTimeoutMilliseconds);
```
RassK, line 89: Should we use `Task.Delay` instead of `Thread.Sleep`? ... due function is anyway async

`await Task.Delay(DefaultTimeoutTimeSpan).ConfigureAwait(true);`

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/KafkaHeadersCollectionAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+                return;
+            }
+
+            TryGetLastBytesMethodInfo = methods.FirstOrDefault(m => m.Name == "TryGetLastBytes");
+            if (TryGetLastBytesMethodInfo == null)
+            {
+                Log.Warning("Missing expected method TryGetLastBytes on type " + ConfluentKafka.HeadersType);
+            }
+
+            AddMethodInfo = methods.FirstOrDefault(m => m.Name == "Add" && m.GetParameters().Length == 2);
```
RassK, line 36: I guess here type safety will be guaranteed with integration tests over all client lib versions from x to y?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/KafkaHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+        public static readonly Lazy<Type> LazyHeadersType = new(() =>
+        {
+            Assembly assembly;
+            try
+            {
+                assembly = Assembly.Load(ConfluentKafka.AssemblyName);
+            }
+            catch (Exception ex)
+            {
+                Log.Error($"Failed to load ${ConfluentKafka.AssemblyName} assembly: {ex.Message}");
```
RassK, line 24: Log should be used with exception overload - specially when catching general exceptions.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/KafkaHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            Type headersType = null;
+            try
+            {
+                headersType = assembly.GetType(ConfluentKafka.HeadersType, throwOnError: true);
+            }
+            catch (Exception ex)
+            {
+                Log.Error($"Failed to get type ${ConfluentKafka.HeadersType}: {ex.Message}");
```
RassK, line 35: Same as above about logging.

src/Datadog.Trace/Tags.cs
pjanotti: Yes, you are correct, however, this is the only thing (I think) that the Confluent.Kafka API has exposing the the client_id so it seems a reasonable to choice to keep as it is.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/KafkaHeadersCollectionAdapter.cs
pjanotti: We have ways to test for all versions but to be fair I only tested 1.4.3 and 1.7.0 (latest release).

samples/Samples.Kafka/Program.cs
pjanotti: Yes, Task.Delay is better (notice that this is in the test application).

src/Datadog.Trace/Tags.cs
dszmigielski: Shouldn't we change it to something along `client_name` then? 🤔

RassK (approved):
Looks promising

pjanotti (approved):
LGTM

PR #128, opened 2021-06-30
Signature fixes: use interface instead of concrete type

pjanotti:
[This is done on top of PR https://github.com/signalfx/signalfx-dotnet-tracing/pull/127]

Two things were blocking the instrumentations:

1. Since this is call site instrumentation you actually need to set the interfaces on the interception attributes. Alternatively one could intercept the builder methods but then return a wrapper to the original implementation that supports the same interface and generate the spans from there (seem very tempting in this case since the API then becomes mostly a pass through, with few exceptions - the same as the methods that we want to intercept).

~~2. The instrumentation on method Main is problematic, see comment added to the sample.~~

Still broken: I had to hard code the value of the timeout on the Consume method since the code is not handling it correctly. I am going to investigate this more.

Item 2 above was a simple mistake: I forgot to regenerate integrations.json when changing the signature of the method.


pjanotti:
@dszmigielski it is probably easier to see the diff with your branch only https://github.com/dszmigielski/signalfx-dotnet-tracing/pull/3

pjanotti:
Closing since this one is not needed anymore (added to #127)

PR #133, opened 2021-07-13
Confluent.Kafka: do not set consumer start time if msg was received

pjanotti:
When a msg is received on the consumer call has its actual start time it distorts the whole trace time by having the consumer span starting before the server receiving the request. This change doesn't use the consumer start time if an actual message is retrieved from the topic.

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/ConsumeKafkaIntegrationHelper.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/ConsumeKafkaIntegrationHelper.cs
@@ -58,39 +58,10 @@ namespace Datadog.Trace.ClrProfiler.Integrations.Kafka
                 throw;
             }
 
-            object result;
-            Scope scope = null;
-            var startTimeOffset = DateTimeOffset.UtcNow;
-            try
+            object result = consume(consumer, input);
+            if (result is not null)
             {
-                result = consume(consumer, input);
-                scope = result != null
-                    ? CreateConsumeScopeFromConsumerResult(result, consumer)
-                    : CreateConsumeScopeFromConsumer(consumer, startTimeOffset);
-            }
-            catch (Exception ex)
-            {
-                var tracer = Tracer.Instance;
-                if (!tracer.Settings.IsIntegrationEnabled(ConfluentKafka.IntegrationName) || KafkaHelper.AlreadyInstrumented())
-                {
-                    throw;
-                }
-
-                // Integration is enabled but the scope was not created since consume raised an exception.
-                // Create a span to record the exception with the available info.
-                scope = CreateConsumeScopeFromConsumer(consumer, startTimeOffset);
-                if (ex is not OperationCanceledException)
-                {
-                    // OperationCanceledException is expected in case of a clean shutdown and shouldn't
-                    // be reported as an exception.
-                    scope.Span.SetException(ex);
-                }
-
-                throw;
-            }
-            finally
-            {
-                scope?.Dispose();
+                using var scope = CreateConsumeScopeFromConsumerResult(result, consumer);
             }
 
             return result;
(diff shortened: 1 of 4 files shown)


src/Datadog.Trace.ClrProfiler.Managed/Integrations/Kafka/ConsumeKafkaIntegrationHelper.cs
```diff
@@ -65,7 +65,7 @@ internal static class ConsumeKafkaIntegrationHelper
             {
                 result = consume(consumer, input);
                 scope = result != null
-                    ? CreateConsumeScopeFromConsumerResult(result, consumer, startTimeOffset)
+                    ? CreateConsumeScopeFromConsumerResult(result, consumer)
                     : CreateConsumeScopeFromConsumer(consumer, startTimeOffset);
```
pellared, line 69: The same issue is here. The `startTimeOffset` is taken from a point before the message was actually retrieved.

pjanotti: In the case when the result is null the desired behavior is to actually record the start time of the call: if there was no data to be consumed, i.e. no error, it records how long the code waited for data and the general feedback was that it was good to as a proof that the consumer was actually working properly.

pellared: Is such behavior documented anywhere or is there some OTel instrumentation that already works that way? 

If I would be a user I would probably hate such behavior, because it would be very inconsistent. The calls that got the message would be extra short (still we do not know how long it was waiting for a message) and we would have some which would be long (probably equal to some polling interval).

E.g. I see that for the Kafka Go client we do not even create a span for such a scenario: https://github.com/signalfx/signalfx-go-tracing/blob/875b403aa2b9f916d74f08a48c04a50f05e8595a/contrib/confluentinc/confluent-kafka-go/kafka/kafka.go#L131-L135

As far as I guess the Java instrumentation also creates spans when messages are being accessed via `record` or `iterator`. I think that are no spans created if e.g. `ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(100));` returns no message. @mateuszrzeszutek can you confirm that, please?

1. We can consider creating a dedicated tag/attribute like `wait_time` to capture a metric of how long the caller was waiting to get a message. 
2. If we decide to create spans for calls that did not receive a message then we could also create an `empty_message` attribute/tag to differentiate between these spans. However, probably it would prefer to NOT create any span in such a case.

It may be good also to define this behavior in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md

mateuszrzeszutek: Don't look at the Java Kafka instrumentation, it's horrible 😂  And causes a lot of issues too - just search for "kafka" in our backlog...
Because it's a poll-only interface we have to use various not-so-pretty hacks to correctly set the context, like wrapping the returned list of records.

> I think that are no spans created if `ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(100));` returns no message. @mateuszrzeszutek can you confirm that, please?

That's correct - if there are no messages no spans will be created at all.

pjanotti: > If I would be a user I would probably hate such behavior, because it would be very inconsistent. The calls that got the message would be extra short (still we do not know how long it was waiting for a message) and we would have some which would be long (probably equal to some polling interval).

I actually suggested removing those spans to the users that tried this instrumentation and they didn't like to remove those spans: they see them as valuable as a "debugging tool": proof that wait was happening on the right topics and partitions. The user didn't feel confused because when it retrieves a message it is a child span of a trace, the ones without messages are root spans. It seems that having it as an option will be the compromise - with the "debug spans" off by default. 

> if we decide to create spans for calls that did not receive a message then we could also create an empty_message attribute/tag to differentiate between these spans.

We already have the tag `"messaging.kafka.message_received": "false"`

I like the idea of `messaging.wait_time`.

pellared: https://docs.confluent.io/platform/current/kafka/monitoring.html#per-topic-metrics can be also used for "debugging tools".  They can use OTel Collector to get these metrics https://docs.signalfx.com/en/latest/integrations/integrations-reference/integrations.kafka.html

pjanotti: The general feedback was to remove the span if no message was received. It is a better case to debug/monitor via metrics or logs (in case of errors).

pellared (approved):
Maybe worth noting the changes in `CHANGELOG.md`?

pellared (approved):

PR #135, opened 2021-08-16
Add Server-Timing header to IIS ASP.NET

pjanotti:
The focus of this PR is to add the `Server-Timing` header to IIS ASP.NET, however, opportunistically it also improves the generation of header in other regards:

- Removed a dead source file;
- Added the error cases that were missing for `Server-Timing` in certain paths;
- Improved some related tests (certain tests are not running on CI, this PR doesn't address that);
- Added component/integration tag to all HTTP server spans;


src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -286,9 +287,16 @@ public static Scope CreateScope(object controllerContext)
                 // call the original method, inspecting (but not catching) any unhandled exceptions
                 var response = instrumentedMethod(asyncControllerActionInvoker, controllerContext, actionName, callback, state);
 
-                if (HttpContext.Current != null && scope != null)
+                if (scope != null)
                 {
-                    ServerTimingHeader.SetHeaders(scope.Span.Context, HttpContext.Current.Response, (resp, name, value) => resp.Headers.Add(name, value));
+                    // TracingHttpModule is expected to already have added it in the typical IIS setup
+                    // In principle we could remove adding the headers from this instrumentation, keeping
+                    // so used can disable "AspNet" instrumentation and still get the Server-Timing header added.
```
RassK, line 294:
```suggestion
                    // so user can disable "AspNet" instrumentation and still get the Server-Timing header added.
```

src/Datadog.Trace/ServerTimingHeader.cs
```diff
@@ -12,7 +12,11 @@ namespace SignalFx.Tracing
     /// </summary>
     public static class ServerTimingHeader
     {
-        private const string ServerTimingHeaderName = "Server-Timing";
+        /// <summary>
+        /// Key of the "Server-Timing" header.
+        /// </summary>
+        public const string Key = "Server-Timing";
```
RassK, line 18: new name `Key` seems to generic. Specially if class already contains 2 "Keys".

johnbley (approved):

dszmigielski (approved):
LGTM

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -165,6 +166,12 @@ private void OnHostingHttpRequestInStart(object arg)
                 Scope scope = _tracer.ActivateSpan(span);
 
                 _options.OnRequest?.Invoke(scope.Span, httpContext);
+
+                var httpContextHeaders = httpContext.Response.Headers;
+                if (httpContextHeaders != null && !httpContextHeaders.ContainsKey(ServerTimingHeader.Key))
```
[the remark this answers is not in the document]
pjanotti: That's a good point. The reason I'm not doing it here is because there are actually two header collection types being handled by this snippet: most places it is dealing with a `NameValueCollection` but some with `IHeaderDictionary` - so in the end I opted to not use an extension method since multiple would be needed.

src/Datadog.Trace/ServerTimingHeader.cs
pjanotti: The name works well when it is used outside of the class, e.g. `httpContextHeaders.ContainsKey(ServerTimingHeader.Key)`, that are dictionary like data types.

PR #142, opened 2021-09-23
Fix null exception on PropagationHelpers

pjanotti:
Quick fix for PropagationHelpers - we have a story to follow up on propagation and will add a test to proper cover this issue (currently there is no test for this case).


RassK (approved):
LGTM

johnbley (approved):

PR #146, opened 2021-09-27
Remove dotnet-tool from minimal CI

pjanotti:
This step is not valuable for our CI at this stage, removing it from minimal CI.


RassK (approved):
LGTM

PR #148, opened 2021-09-28
next-ver 1st upstream pull

pjanotti:
First pull from upstream on next-ver branch.

Last SHA from upstream cecac4ffd.

The instruction from OpenTelemetry work pretty well the only recommendation not there is to not change the default pipeline from upstream and use our own for better fine control.


.azure-pipelines/steps/run-in-docker.yml
```diff
@@ -31,6 +31,8 @@ steps:
         --env tracerHome=/project/$(relativeTracerHome) \
         --env artifacts=/project/$(relativeArtifacts) \
         --env SIGNALFX_CLR_ENABLE_NGEN=$(SIGNALFX_CLR_ENABLE_NGEN) \
+        --env Verify_DisableClipboard=true \
+        --env DiffEngine_Disabled=true \
```
pellared, line 35: this should be added to our pipeline

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -54,21 +54,24 @@ schedules:
...
+  isPullRequest: $[eq(variables['Build.Reason'], 'PullRequest')]
   DD_DOTNET_TRACER_MSBUILD:
   NugetPackageDirectory: $(System.DefaultWorkingDirectory)/packages
   relativeNugetPackageDirectory: packages
   # For scheduled builds, only run benchmarks and crank (and deps).
   isScheduledBuild: ${{ eq(variables['Build.Reason'], 'Schedule') }} # only works if you have a main branch
   # Allow forcing a benchmark/throughput run using run_benchmarks_only=true
   dotnetToolTag: build-dotnet-tool
+  Verify_DisableClipboard: true
+  DiffEngine_Disabled: true
```
pellared, line 74: this should be added to our pipeline

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -410,7 +413,7 @@ stages:
 
     - publish: tracer/build_data
       displayName: Uploading integration_tests_windows tracer logs
-      artifact: integration_tests_windows_tracer_logs_$(platform)_$(target)
+      artifact: integration_tests_windows_tracer_logs_$(platform)_$(target)_$(System.JobAttempt)
```
pellared, line 416: this should be added to our pipeline

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -54,21 +54,24 @@ schedules:
 # Global variables
 variables:
   buildConfiguration: Release
-  dotnetCoreSdk5Version: 5.0.103
+  dotnetCoreSdk5Version: 5.0.401
```
pellared, line 57: this should be added to our pipeline

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -160,6 +160,15 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
                          " is recognized as Kudu, an Azure App Services reserved process.");
             return E_FAIL;
         }
+
+        const auto functions_worker_runtime_value =
+            GetEnvironmentValue(environment::azure_app_services_functions_worker_runtime);
+
+        if (!functions_worker_runtime_value.empty() && !IsAzureFunctionsEnabled())
+        {
+            Logger::Info("DATADOG TRACER DIAGNOSTICS - Profiler disabled: Azure Functions are not officially supported. Enable instrumentation with SIGNALFX_TRACE_AZURE_FUNCTIONS_ENABLED.");
```
pellared, line 169: Should we rename `DATADOG` to `SIGNALFX` here?

.azure-pipelines/steps/run-in-docker.yml
pjanotti: Right, we need to add that for the GH workflow. For the minimal pipeline we get that indirectly since it also uses the run-in-docker file

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
pjanotti: Let's do logs as a whole instead of piece by piece.

pellared (approved):

PR #149, opened 2021-09-28
Remove alpine CI jobs

pellared:
## What

Remove alpine CI jobs.

## Why

Decrease non-critical jobs to reduce the resource usage on Azure Pieplines.


pjanotti (approved):
LGTM. Since you have to rebase anyway, can you also remove Windows x86 tests for the time being (build and unit tests fine but remove regression and integration).

pjanotti:
Also disable code coverage for the time being since we don't have a report for it yet in place.

pellared:
@pjanotti You need to loosen the branch protection

image

PR #150, opened 2021-09-29
Update dev/envvars.sh

pellared:
## Why

The build locations have changed.

## What 

Update `dev/envvars.sh` to facilitate manual testing.

### DIFF
--- a/dev/envvars.sh
+++ b/dev/envvars.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -euo pipefail
 
 uname_os() {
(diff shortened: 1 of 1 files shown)


dev/envvars.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -euo pipefail
```
pellared, line 2: AFAIK scripts that are intended to be used like `source script.sh` should not do such things as it affects the current terminal session

dszmigielski (approved):

pjanotti (approved):

PR #156, opened 2021-10-05
Limit zipkin tag value length

pjanotti:
Implement tag truncation per current implementation: truncate tag values at export time.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
@@ -393,6 +393,14 @@ namespace Datadog.Trace.Configuration
         public const string KafkaCreateConsumerScopeEnabled = "SIGNALFX_TRACE_KAFKA_CREATE_CONSUMER_SCOPE_ENABLED";
 
         /// <summary>
+        /// Configuration key to set maximum length a tag/log value can have.
+        /// Values are completely truncated when set to 0, and ignored when set to negative
+        /// or non-integer string. The default value is 1200.
+        /// </summary>
+        /// <seealso cref="TracerSettings.RecordedValueMaxLength"/>
+        public const string RecordedValueMaxLength = "SIGNALFX_RECORDED_VALUE_MAX_LENGTH";
+
+        /// <summary>
         /// String format patterns used to match integration-specific configuration keys.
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -33,6 +33,8 @@ public class TracerSettings
         /// </summary>
         public const int DefaultAgentPort = 8126;
 
+        private const int DefaultRecordedValueMaxLength = 1200;
```
pellared, line 36: The default should be `12000`.
See: https://github.com/signalfx/gdi-specification/pull/130

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -392,6 +392,14 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.KafkaCreateConsumerScopeEnabled"/>
         public const string KafkaCreateConsumerScopeEnabled = "SIGNALFX_TRACE_KAFKA_CREATE_CONSUMER_SCOPE_ENABLED";
 
+        /// <summary>
+        /// Configuration key to set maximum length a tag/log value can have.
+        /// Values are completely truncated when set to 0, and ignored when set to negative
+        /// or non-integer string. The default value is 1200.
+        /// </summary>
+        /// <seealso cref="TracerSettings.RecordedValueMaxLength"/>
+        public const string RecordedValueMaxLength = "SIGNALFX_RECORDED_VALUE_MAX_LENGTH";
```
pellared, line 401: Can you document it in `USAGE.md`?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
pellared: Addressed 35138ef6e

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
pellared: Addressed bcc865a3c1a8

pellared (approved):

RassK (approved):
LGTM

PR #157, opened 2021-10-05
Support SIGNALFX_ACCESS_TOKEN env var

pellared:
Add `SIGNALFX_ACCESS_TOKEN`  which enables sending traces to a Splunk Observability Cloud ingest endpoint.

### DIFF
--- a/tracer/src/Datadog.Trace/Agent/Zipkin/ZipkinExporter.cs
+++ b/tracer/src/Datadog.Trace/Agent/Zipkin/ZipkinExporter.cs
@@ -1,3 +1,5 @@
+// Modified by Splunk Inc.
+
 using System;
 using System.Net;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


pellared:
@pjanotti I assume that in order to test it I must also configure `SIGNALFX_TRACE_AGENT_URL`. What is the proper value that can be used as I cannot find info even on the `main` branch?

tracer/src/Datadog.Trace/Agent/Zipkin/ZipkinExporter.cs
```diff
@@ -45,6 +47,12 @@ public async Task<bool> SendTracesAsync(Span[][] traces)
                 // Disable automatic instrumentation for Zipkin exporter
                 request.Headers.Add(CommonHttpHeaderNames.TracingEnabled, "false");
 
+                // Add SignalFx Access Token if configured
+                if (!string.IsNullOrWhiteSpace(_settings.SignalFxAccessToken))
```
RassK, line 51: nit: could use read only precalculated setting 

`bool IsAccessTokenSet = string.IsNullOrWhiteSpace(SignalFxAccessToken)`

so due `SendTraces` is executed constantly, it will be faster.... maybe a ms diff 😄

RassK (approved):
LGTM

tracer/src/Datadog.Trace/Agent/Zipkin/ZipkinExporter.cs
pellared: you are right... however, it would probably consume a little more memory 😄 

_for sure the current code is a little more readable_ 😉

pjanotti (approved):
LGTM

PR #163, opened 2021-10-10
Update HttpMessageHandler semantic conventions

pjanotti:
The difference between SFx and upstream is the span name: on SFx (similar to OTel) we use the HTTP method name (in upper case), while upstream uses the fixed value of “http.request”. Notice that the ResourceName used by upstream in this case is very close to the name that we want but since it adds the path to the resource name it isn’t a good candidate for the span name due to cardinality.


tracer/src/Datadog.Trace/ClrProfiler/ScopeFactory.cs
```diff
@@ -120,15 +122,38 @@ internal static Scope CreateOutboundHttpScope(Tracer tracer, string httpMethod,
...
+                string resourceUrl = requestUri != null ? UriHelpers.CleanUri(requestUri, removeScheme: true, tryRemoveIds: true) : null;
+                string httpUrl = requestUri != null ? UriHelpers.CleanUri(requestUri, removeScheme: false, tryRemoveIds: false) : null;
+
+                tags = new HttpTags();
+
+                // Upstream uses "http.request" for the span name but following legacy version and the
+                // OpenTelemetry specification we use the capitalized method as the span name.
+                string uppercaseHttpMethod = httpMethod?.ToUpperInvariant();
+
+                string serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
```
pjanotti, line 134: This adds a suffix to the service name, in this case "-http-client". At first my thought was to address this by updating the helper function. However, I'm starting to think that it will be better to actually do it in this change: so we change affected for each instrumentation affected when we update the instrumentation.

pellared:
`// Modified by Splunk Inc.` is missing in some files.

I think the same issue might be in the previous merged PR.

pellared (approved):

pjanotti:
@pellared, you're correct we are missing the added header in a few places. I'm looking for a way to automate this process.

PR #164, opened 2021-10-10
Fix tag used for environment

pjanotti:
The tag "env", used by upstream, is not recognized by the backend.
Legacy uses "environment" and while that is supported by the backend
it requires use of global tags when combining with k8s data collection.
Using the current recommendation from GDI spec.


pjanotti:
I didn't change the usage of "env" in metrics because we are not exporting metrics at this moment.

pellared (approved):

PR #165, opened 2021-10-12
Propagation optimizations for W3C

dszmigielski:
**What**
Reapply 'Performance pitfall in W3C propagation' changes. In the process I have removed the DDand Otel convention setting, so we are now always using Otel TraceId convention. If that is not what we want in the end please comment :)

**Why**
The changes were overridden by the upstream pull


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -329,14 +329,6 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.Exporter"/>
         public const string Exporter = "SIGNALFX_EXPORTER";
 
-        /// <summary>
-        /// Configuration key for the semantic convention to be used.
-        /// The Tracer uses it to define operation names, span tags, statuses etc.
-        /// Default is <c>"Default"</c>.
-        /// <seealso cref="TracerSettings.Convention"/>
-        /// </summary>
-        public const string Convention = "SIGNALFX_CONVENTION";
```
pellared, line 338: please remove `SIGNALFX_CONVENTION` from everywhere (e.g. docs, scripts, etc)

dszmigielski: Thanks for noting that, I remembered that when I was beginning the removal process...

tracer/src/Datadog.Trace/Tracer.cs
```diff
@@ -125,16 +125,7 @@ internal Tracer(TracerSettings settings, IReadOnlyCollection<IOTelExtension> plu
...
-            {
-                case ConventionType.OpenTelemetry:
-                    TraceIdConvention = new OtelTraceIdConvention();
-                    break;
-                case ConventionType.Datadog:
-                default:
-                    TraceIdConvention = new DatadogTraceIdConvention();
-                    break;
-            }
+            TraceIdConvention = new OtelTraceIdConvention();
```
pjanotti, line 128: Look at how upstream handle this, likely the convention property can be removed since the attempt is to use a single one on the whole code base. However, as mentioned before it probably better addressed in a dedicated PR.

dszmigielski: Yep, actually I started to do that because of the TraceId changes that we then dropped. So this is not required for now at all. Reverted.

pellared (approved):

PR #166, opened 2021-10-13
Update semantic conventions for TracingHttpModule

pjanotti:
__Not ready yet__ just checking test impact of current changes.


tracer/src/Datadog.Trace/AspNet/TracingHttpModule.cs
```diff
@@ -123,12 +123,13 @@ private void OnBeginRequest(object sender, EventArgs eventArgs)
 
                 string host = httpRequest.Headers.Get("Host");
                 string httpMethod = httpRequest.HttpMethod.ToUpperInvariant();
-                string url = httpRequest.RawUrl.ToLowerInvariant();
+                string url = httpRequest.Url.ToString(); // Upstream uses RawUrl, ie. the part of the URL following the domain information.
```
RassK, line 126: Shouldn't this be also lowered?

RassK: isn't it going to break cardinality?

RassK: ... although I would expect the portal (back-end) to handle it.

pjanotti: Let's continue this conversation via Slack and talk with some backend folks. If needed we revisit it.

PR #167, opened 2021-10-13
Fix RabbitMq delivery mode tags

RassK:
(bug) RabbitMq delivery mode tags got swapped with RouteKey, which caused duplicate tags and crashed client application.

### DIFF
--- a/samples/Samples.RabbitMQ/Program.cs
+++ b/samples/Samples.RabbitMQ/Program.cs
@@ -74,7 +74,7 @@ namespace Samples.RabbitMQ
 
                     channel.BasicPublish(exchange: exchangeName,
                                             routingKey: routingKey,
-                                            basicProperties: null,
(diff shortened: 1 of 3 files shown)


pellared:
Why/How was it crashing? Can a tag be duplicated? Is it not a simple dictionary so the new value is replacing the old one?

RassK:
> Why and how was it crashing? Can a tag be duplicated? Is it not a simple dictionary so the new value is replacing the old one?

main branch has old simplified tag model, just a common dictionary.

pellared (approved):
I see. Thanks

SIde note. Maybe it is better to use `tags[key] = ...` instead of `tags.Add(key, ...` to minimize the chance of throwing an exception and have probably a little better performance.

pjanotti:
> Actually it should not have been the client's app that crashed but our integration tests.

@RassK can we add such a test or is it the cost too high?

PR #169, opened 2021-10-14
Minimal GitHub workflow - Windows jobs

pellared:
Migrate Windows jobs from `mininal-pipeline.yml`  to GH Actions.

## Next PRs

- Handle Linux jobs. I consider to use "composite actions" to handle `steps/run-in-docker.yml`. More: https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md or just run everything on Ubuntu

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -87,8 +87,6 @@ jobs:
   package_windows_Stage_msi_and_pack:
     name: Windows - msi and nuget
     runs-on: windows-2019
-    needs:
-    - build_windows_Stage_build
     steps:
     - uses: actions/checkout@v2
    # - # There is no conversion path for templates in GitHub Actions
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -84,37 +84,6 @@ jobs:
   #     with:
   #       path: ${{ env.binDir }}
   #       name: build-linux-${{ env.baseImage }}-working-directory
-  package_windows_Stage_msi_and_pack:
-    name: Windows - msi and nuget
-    runs-on: windows-2019
-    steps:
-    - uses: actions/checkout@v2
-   # - # There is no conversion path for templates in GitHub Actions
-   #   run: '#steps/install-dotnet-5-sdk.yml'
-    - name: Build tracer home
-      run: tracer\build.cmd BuildTracerHome
-    - name: Build MSI and Tracer home
-      run: tracer\build.cmd PackageTracerHome
-    - name: Publish tracer-home.zip
-      uses: actions/upload-artifact@v2
-      with:
-        path: ${{ env.artifacts }}/windows-tracer-home.zip
-        name: windows-tracer-home.zip
-    - name: Publish Windows x86 MSI
-      uses: actions/upload-artifact@v2
-      with:
-        path: ${{ env.artifacts }}/x86/en-us
-        name: windows-msi-x86
-    - name: Publish Windows x64 MSI
-      uses: actions/upload-artifact@v2
-      with:
-        path: ${{ env.artifacts }}/x64/en-us
-        name: windows-msi-x64
-    - name: Publish NuGet packages
-      uses: actions/upload-artifact@v2
-      with:
-        path: ${{ env.artifacts }}/nuget
-        name: nuget-packages
   unit_tests_windows_Stage_managed:
     name: Windows - managed unit tests
     runs-on: windows-2019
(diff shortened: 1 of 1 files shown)


pellared:
Someone can rerun the Azure Pipeline job please?

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
+name: minimal
+on:
+  push:
+    branches:
+    - main
+    - next-ver
+    tags:
+    - 'v*'
+    paths-ignore:
```
pjanotti, line 9: 🤞 that this will work :) we tried something like this in AZ pipelines and it didn't 🤞

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+    - docs/*
+    - shared/samples/Datadog.AutoInstrumentation.ManagedLoader.Demo/*
```
pjanotti, line 11: Could it instead be `shared/samples/**/*`, or the equivalent, instead of enumeration different samples? similar to others. BTW, does the traces actually uses some of these shared stuff?

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+    - shared/src/native-lib/fmt_x64-windows-static/*
+    - shared/src/native-lib/fmt_x86-windows-static/*
+    - shared/src/native-lib/spdlog/*
+    - shared/src/native-src/*
+  pull_request:
+  workflow_dispatch:
+env:
+  buildConfiguration: Release
+  dotnetCoreSdk5Version: 5.0.401
+  relativeTracerHome: /tracer/src/bin/windows-tracer-home
```
pjanotti, line 30: the fact that it starts with `/` makes it look like an absolute path. Similar to the next one.

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+  relativeArtifacts: /tracer/src/bin/artifacts
+  binDir: ${{ github.workspace }}/tracer/src/bin
+  ddTracerHome: ${{ github.workspace }}/tracer/src/bin/dd-tracer-home
```
pjanotti, line 33: I've been using the tracer-home do you what is the purpose of the dd one?

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+  #     uses: actions/upload-artifact@v2
+  #     with:
+  #       path: ${{ env.artifacts }}/linux-x64
+  #       name: linux-packages-${{ env.baseImage }}
+  #   - name: Upload working directory after the build
+  #     uses: actions/upload-artifact@v2
+  #     with:
+  #       path: ${{ env.binDir }}
+  #       name: build-linux-${{ env.baseImage }}-working-directory
+  unit_tests_windows_Stage_managed:
```
pjanotti, line 87: nit: 
```suggestion
  unit_tests_windows_stage_managed:
```

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+  tracerHome: ${{ github.workspace }}/tracer/src/bin/windows-tracer-home
+  artifacts: ${{ github.workspace }}/tracer/src/bin/artifacts
+  isMainBranch: $[eq(github.ref, 'refs/heads/next-ver')]
+  NugetPackageDirectory: ${{ github.workspace }}/packages
+  relativeNugetPackageDirectory: packages
+  dotnetToolTag: build-dotnet-tool
+  Verify_DisableClipboard: true
+  DiffEngine_Disabled: true
+jobs:
+  build_windows_Stage_build:
```
pjanotti, line 43: nit: 
```suggestion
  build_windows_stage_build:
```

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+  #       name: build-linux-${{ env.baseImage }}-working-directory
+  unit_tests_windows_Stage_managed:
+    name: Windows - managed unit tests
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+   # - # There is no conversion path for templates in GitHub Actions
+   #   run: '#steps/install-dotnet.yml'
+    - name: Build and Test
+      run: tracer\build.cmd BuildTracerHome BuildAndRunManagedUnitTests
```
pjanotti, line 95: Is it possible to use the artifacts from the other job? Or collect the artifacts from this one instead?

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+    - uses: actions/upload-artifact@v2
+      with:
+        path: tracer/build_data
+        name: profiler-logs_unit_tests_windows_${{ env.System.JobAttempt }}
+      if: (${{ job.status }} != 'cancelled')
+      continue-on-error: true
+  unit_tests_windows_Stage_native:
```
pjanotti, line 102: 🤔 hum starting to thinking that the capitalized `Stage` is on purpose, is that the case?

.github/workflows/minimal.yml
```diff
@@ -0,0 +1,250 @@
...
+      run: tracer\build.cmd BuildTracerHome
+    - name: Start CosmosDB Emulator
+      run: |
+        Write-Host "Starting CosmosDB Emulator"
+        Import-Module "C:/Program Files/Azure Cosmos DB Emulator/PSModules/Microsoft.Azure.CosmosDB.Emulator"
+        Start-CosmosDbEmulator -Timeout 300
+      shell: powershell
+      if: ${{ matrix.requiresCosmos }}
+    - name: Run integration tests
+      run: tracer\build.cmd ${{ matrix.target }} -Framework netcoreapp3.1 --PrintDriveSpace
```
pjanotti, line 163: While x64 is the default I think it will be clear if we pass the matrix Platform to keep it consistent.

.github/workflows/minimal.yml
pellared: I will put the comments from https://github.com/DataDog/dd-trace-dotnet/blob/master/.azure-pipelines/ultimate-pipeline.yml#L15

.github/workflows/minimal.yml
pellared: I want to clean it in the next PR. The names come from automatic conversion done via https://pipelinestoactions.azurewebsites.net/

.github/workflows/minimal.yml
pellared: Yes, it is. I can check it later but I basically migrated the existing pipeline and kept the matrix where the matrixes where there were used in Azure Pipelines

pellared: Now I see that the upstream set TargetPlatform via env var. I am fixing this via CLI arg instead (more readable)

dszmigielski (approved):

PR #170, opened 2021-10-15
Enhance the GitHub Actions workflow

pellared:
## What

- Add .NET 5.0 for integration tests jobs
- Add Ubuntu build and unit test jobs
- Add Debian container integration tests jobs

26 minutes for all 13 jobs 🎉 
compared to 40 min NET (not including queuing/waiting which can sometimes take even hours) in Azure Pipelines (8 jobs)

## Next PR

- Investigate issues with integration test on .NET Framework
- Investigate issues with integration test on Alpine (probably out of scope)

### Install SQL Express on Ubuntu

We get `Unhandled exception. System.PlatformNotSupportedException: LocalDB is not supported on this platform.` when I tried running integration tests on "bare-metal" Ubuntu.

We can install SQL Express on Ubuntu. More:
- https://docs.microsoft.com/en-us/sql/linux/quickstart-install-connect-ubuntu?view=sql-server-ver15
- https://docs.microsoft.com/en-us/sql/linux/sample-unattended-install-ubuntu?view=sql-server-ver15

However, I decided just to reuse the existing docker-compose approach so it can be tested inside Debian and Alpine docker containers.

I still wanted to put this comment if someone would like to run the integration tests on WSL without using `docker-compose IntegrationTests`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,7 +196,7 @@ jobs:
           IntegrationTests
     - uses: actions/upload-artifact@v2
       with:
-        name: ${{ matrix.machine }}-integration-tests-${{ matrix.platform }}-${{ matrix.framework }}-${{ matrix.target }}-build_data
+        name: ${{ matrix.base-image }}-integration-tests-${{ matrix.framework }}-build_data
         path: tracer/build_data
       if: (${{ job.status }} != 'cancelled')
       continue-on-error: true


.github/workflows/minimal.yml
```diff
@@ -53,109 +53,96 @@ env:
...
+  integration-tests:
+    name: Integration tests 
     strategy:
       fail-fast: false
       matrix:
         include:
-        - platform: x64
+        - machine: windows-2019
+          platform: x64
+          framework: netcoreapp3.1
```
pjanotti, line 127: we should have a Fx (net46?) for the integration tests.

.github/workflows/minimal.yml
```diff
@@ -1,245 +0,0 @@
...
-  #   - # There is no conversion path for templates in GitHub Actions
-  #     run: |
-  #       #steps/run-in-docker.yml
-  #       build: true
-  #       baseImage: ${{ env.baseImage }}
-  #       command: BuildTracerHome BuildLinuxIntegrationTests --framework ${{ env.publishTargetFramework }}
-  #   - name: docker-compose build IntegrationTests
-  #     run: docker-compose build --build-arg baseImage=${{ env.baseImage }} --build-arg framework=${{ env.publishTargetFramework }} IntegrationTests
-  #   - name: docker-compose start dependencies
-  #     run: docker-compose run --rm StartDependencies
```
pellared, line 233: I might have missed to add this line in the case of Linux integration tests...

.github/workflows/minimal.yml
```diff
@@ -53,109 +53,96 @@ env:
...
-  #   - uses: actions/upload-artifact@v2
-  #     with:
-  #       path: tracer/build_data
-  #       name: profiler-logs_unit_tests_linux_${{ env.Agent.JobName }}_${{ env.System.JobAttempt }}
-  #     if: (${{ job.status }} != 'cancelled')
-  #     continue-on-error: true
-  windows-integration-tests:
-    name: Windows - integration tests 
-    runs-on: windows-2019
+    - run: ./tracer/build.cmd BuildTracerHome BuildAndRunNativeUnitTests
```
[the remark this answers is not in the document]
pellared: Especially that we may not notice when the upstream implements it as they also run the unit tests in their pipeline.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,229 @@
...
+          --env tracerHome=/project/${relativeTracerHome} \
+          --env artifacts=/project/${relativeArtifacts} \
+          --env SIGNALFX_CLR_ENABLE_NGEN=${SIGNALFX_CLR_ENABLE_NGEN} \
+          dd-trace-dotnet/${{ matrix.base-image }}-builder \
+          dotnet /build/bin/Debug/_build.dll BuildTracerHome BuildLinuxIntegrationTests -Framework ${{ matrix.framework }}
+    - name: Start dependencies
+      run: docker-compose run --rm StartDependencies
+    - name: Run integration tests in container
+      run: | 
+        framework=${{ matrix.framework }} docker-compose run --rm \
```
pjanotti, line 191: Why is the `framework` env var set for for `docker-compose run`? Notice that is also injected in the docker container directly on the command to start the container.

pjanotti (commented):
Why is tracer/build.cmd showing as an empty file?

pjanotti:
@pellared I'm looking at some of the test failures for net45 and so far I found a few tests that need to be updated (we weren't running them on AZ minimal pipeline). There was also one failure to regarding maximum path exceeded:

```
Starting code execution in AppDomain Samples.MultiDomainHost.App.FrameworkHttpNoRedirects
   Error: System.IO.FileLoadException: Could not load file or assembly 'Samples.MultiDomainHost.App.FrameworkHttpNoRedirects' or one of its dependencies. The filename or extension is too long.
```

I think it will be better to merge without those tests enabled and we can work concurrently to fix them.

.github/workflows/ci.yml
pellared: It was a tricky one.

The env set only "changes" this line: https://github.com/pellared/signalfx-dotnet-tracing/blob/gh-ubuntu/docker-compose.yml#L297

But it occurs that this line: https://github.com/pellared/signalfx-dotnet-tracing/blob/gh-ubuntu/docker-compose.yml#L290 uses host's env vars.

More: https://docs.docker.com/compose/environment-variables/

pellared:
> Why is tracer/build.cmd showing as an empty file?

Execute permission was added so that it can be used by MacOS and Linux 😉

PR #171, opened 2021-10-18
Fix b3 propagation exception

dszmigielski:
Fix bug in B3 propagation

```
System.ArgumentNullException: Value cannot be null.
Parameter name: source
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Datadog.Trace.Propagation.B3SpanContextPropagator.ParseB3Sampling[T](T carrier, Func`3 getter)
   at Datadog.Trace.Propagation.B3SpanContextPropagator.Extract[T](T carrier, Func`3 getter)
   at Datadog.Trace.Propagation.CompositeTextMapPropagator.Extract[T](T carrier, Func`3 getter)
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.Http.WebRequest.HttpWebRequest_BeginGetResponse_Integration.OnMethodBegin[TTarget](TTarget instance, AsyncCallback callback, Object state)
   at HttpWebRequest_BeginGetResponse_Integration.OnMethodBegin(HttpWebRequest , AsyncCallback , Object )
   at Datadog.Trace.ClrProfiler.CallTarget.Handlers.BeginMethodHandler`4.Invoke(TTarget instance, TArg1 arg1, TArg2 arg2)
   at System.Net.HttpWebRequest.BeginGetResponse(AsyncCallback callback, Object state)
```

### DIFF
--- a/tracer/src/Datadog.Trace/Propagation/B3SpanContextPropagator.cs
+++ b/tracer/src/Datadog.Trace/Propagation/B3SpanContextPropagator.cs
@@ -99,14 +99,14 @@ namespace Datadog.Trace.Propagation
 
         private static SamplingPriority? ParseB3Sampling<T>(T carrier, Func<T, string, IEnumerable<string>> getter)
         {
-            var debugged = getter(carrier, B3HttpHeaderNames.B3Flags).ToList();
-            var sampled = getter(carrier, B3HttpHeaderNames.B3Sampled).ToList();
+            var debugged = getter(carrier, B3HttpHeaderNames.B3Flags)?.ToList();
+            var sampled = getter(carrier, B3HttpHeaderNames.B3Sampled)?.ToList();
 
-            if (debugged.Count != 0 && (debugged[0] == "0" || debugged[0] == "1"))
+            if (debugged != null && debugged.Count != 0 && (debugged[0] == "0" || debugged[0] == "1"))
             {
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/Propagation/B3SpanContextPropagator.cs
```diff
@@ -99,14 +99,14 @@ private static ulong ParseHexUInt64<T>(T carrier, Func<T, string, IEnumerable<st
 
         private static SamplingPriority? ParseB3Sampling<T>(T carrier, Func<T, string, IEnumerable<string>> getter)
         {
-            var debugged = getter(carrier, B3HttpHeaderNames.B3Flags).ToList();
-            var sampled = getter(carrier, B3HttpHeaderNames.B3Sampled).ToList();
+            var debugged = getter(carrier, B3HttpHeaderNames.B3Flags)?.ToList();
+            var sampled = getter(carrier, B3HttpHeaderNames.B3Sampled)?.ToList();
```
pellared, line 103: side comment: while the current code fixes the bug I think the code does unnecessary heap allocation because of `ToList` and in general LINQ usage (in the case when B3 is used and should be sampled)

however I think this PR is good and we should not address performance in this PR (maybe it is even not an issue)

PR #172, opened 2021-10-19
Update rabbitmq semantics

RassK:
Changing current tags according to Otel semantics.

I try to do it in 3 steps:
1. Replace existing information & Fix integration tests
2. Add missing information that is easily available


tracer/src/Datadog.Trace/ClrProfiler/Integrations/RabbitMQIntegration.cs
```diff
@@ -855,18 +865,36 @@ internal static Scope CreateScope(Tracer tracer, out RabbitMQTags tags, string c
...
+                // Network tags
+                // TODO: tags.PeerName = "";
+                // TODO: tags.PeerIP = "";
+
+                // Messaging Tags
+                tags.System = ServiceName;
+                tags.Destination = exchange;
+                tags.DestinationKind = SpanTypes.Queue;
+                // TODO: tags.TempDestination = "TODO";
+                // TODO: tags.Protocol = "TODO";
```
RassK, line 883: These 3 can be fetched from Model > Session

RassK:
Any ideas what to do with these exceptional instrumentations that are not defined in OTel?
`QueueBind`
`QueueDeclare`
`ExchangeDeclare`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
```diff
@@ -85,16 +87,18 @@ public static CallTargetState OnMethodBegin<TTarget>(TTarget instance, string qu
...
             }
 
-            using (var scope = RabbitMQIntegration.CreateScope(tracer, out RabbitMQTags tags, Command, parentContext: propagatedContext, spanKind: SpanKinds.Consumer, queue: queue, startTime: startTime))
+            using (var scope = RabbitMQIntegration.CreateScope(tracer, out RabbitMQTags tags, Operation, parentContext: propagatedContext, spanKind: SpanKinds.Consumer, queue: queue, startTime: startTime))
             {
                 if (scope != null)
                 {
                     string queueDisplayName = string.IsNullOrEmpty(queue) || !queue.StartsWith("amq.gen-") ? queue : "<generated>";
-                    scope.Span.ResourceName = $"{Command} {queueDisplayName}";
+                    scope.Span.ResourceName = $"{Operation} {queueDisplayName}";
```
pjanotti, line 95: Doesn't this also works better as the name of the span? Also at this point we are not exporting the resource name since most of the time, so far, it is redundant with span name and/or other tags.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/RabbitMQIntegration.cs
```diff
@@ -899,26 +933,6 @@ public struct BasicGetResultStruct
             public IBasicProperties BasicProperties;
         }
 
-        public interface IBasicProperties
```
pjanotti, line 902: Why this is not needed anymore?

pjanotti (commented):
Still WIP but some initial Qs.

I expected to see the LogicScope set with what was the span name (field OperationName) set by upstream. Why is that?

For review purposes it will be good to compare main spans with spans from this PR.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/RabbitMQIntegration.cs
```diff
@@ -855,18 +865,36 @@ internal static Scope CreateScope(Tracer tracer, out RabbitMQTags tags, string c
...
+                // TODO: tags.Protocol = "TODO";
+                // TODO: tags.ProtocolVersion = "TODO";
+                // TODO: tags.Url = "TODO";
+
+                if (operation == OperationReceive ||
+                    operation == OperationProcess)
+                {
+                    tags.Operation = operation;
+
+                    // tags.ConsumerId = "TODO";
```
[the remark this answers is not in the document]
pjanotti: I don't think we did this on main branch. Unless it is cheaply available we should ignore it: we don't need to extend what already exists in main.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/RabbitMQIntegration.cs
RassK: duplicate interface, I reused the same defined separately in CallTarget folder.

Datadog.Trace.sln
```diff
@@ -341,7 +341,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DuplicateTypeProxy", "trace
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.AspNetCoreRazorPages", "tracer\test\test-applications\integrations\Samples.AspNetCoreRazorPages\Samples.AspNetCoreRazorPages.csproj", "{1B9E6BF4-9D48-4988-9945-248096119E46}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.Vendoring", "tracer\test\test-applications\integrations\Samples.Vendoring\Samples.Vendoring.csproj", "{0D4ACA4A-44DC-4603-8992-926B5BED5D76}"
```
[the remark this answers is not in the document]
RassK: Seems VS doing an autochange on load. I guess it has old project type guid, that VS is trying to update. (all others have the same one, that VS changed to).

RassK:
> Still WIP but some initial Qs.
> 
> I expected to see the LogicScope set with what was the span name (field OperationName) set by upstream. Why is that?
> 
> For review purposes it will be good to compare main spans with spans from this PR.

Yep, it's in very beginning stage. Wanted to get some early feedback, how much OTel we should go as this is lacking a lot in spec wise and if the additions are ok.

PR #173, opened 2021-10-20
Adjust ado.net based integrations to Otel spec

lachmatt:
Adjusts ADO.NET based integrations to Otel spec

@signalfx/instrumentation

### DIFF
--- a/tracer/src/Datadog.Trace/Tags.cs
+++ b/tracer/src/Datadog.Trace/Tags.cs
@@ -88,7 +88,11 @@ namespace Datadog.Trace
         /// <summary>
         /// The type of database (e.g. mssql, mysql)
         /// </summary>
-        public const string DbType = "db.type";
+        /// /// <remarks>
+        /// Upstream uses "db.type", however, to better align with OpenTelemetry we use
+        /// "db.system" instead.
+        /// </remarks>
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Trace/Tags.cs
```diff
@@ -88,7 +88,11 @@ public static class Tags
         /// <summary>
         /// The type of database (e.g. mssql, mysql)
         /// </summary>
-        public const string DbType = "db.type";
+        /// <remarks>
+        /// Upstream uses "db.type", however, to better align with OpenTelemetry we use
+        /// "db.system" instead.
+        /// </remarks>
+        public const string DbType = "db.system";
```
pjanotti, line 95: We are not using in our tests yet but as a fyi there are files tracking the public API that will be affected by this. In principle I think that it is a good idea to do this same update there, look at the tracer\test\Datadog.Trace.Tests\snapshots\PublicApiTests.*.verified.txt files

pjanotti (approved):

pellared:
> the span name is not being changed

I spend together with @lachmatt about an hour discussing the span name. And we decided to keep it as it is as this is even what the `main` branch has.

PR #188, opened 2021-11-01
Update build assets

pjanotti:
Changed the name and branding of build assets to match what we currently have on the main branch. They are already published via GH actions (eg.: https://github.com/signalfx/signalfx-dotnet-tracing/actions/runs/1409408003).

We still need to automate the release publication and add an Alpine musl.tar.gz - these will be done in a separate PR.


RassK (approved):
LGTM

dszmigielski (approved):

PR #190, opened 2021-11-02
[WIP] Update mongodb semantics

RassK:
Updates mongodb semantics


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -437,6 +437,13 @@ public static class ConfigurationKeys
         /// </summary>
         public const string CIVisibilityEnabled = "SIGNALFX_CIVISIBILITY_ENABLED";
 
+        /// <summary>
+        /// Configuration key for enabling or disabling the tagging of
+        /// a Mongo command BsonDocument as db.statement.
```
pjanotti, line 442: nit: mention that the default is `true`.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/MongoDbTests.cs
```diff
@@ -19,28 +27,37 @@ public MongoDbTests(ITestOutputHelper output)
...
             }
         }
 
         [SkippableTheory]
         [MemberData(nameof(GetMongoDb))]
         [Trait("Category", "EndToEnd")]
-        public void SubmitsTraces(string packageVersion, bool enableCallTarget)
+        public void SubmitsTraces(string packageVersion, bool enableCallTarget, bool tagCommands)
         {
+            SetEnvironmentVariable("SIGNALFX_INSTRUMENTATION_MONGODB_TAG_COMMANDS", tagCommands.ToString().ToLowerInvariant());
```
pjanotti, line 46: nit: it should reset the env var at the end of the test

pjanotti (commented):
@RassK we should also remove the spans for "Heartbeat" or "RoundTripTimeMonitor", see https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/207/files

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/MongoDbTests.cs
pjanotti: In practice it shouldn't be a problem but just to keep it clean. Doing it at a tear down for the MongoDb tests also ensures the clean-up.

RassK: Seems clean up is done before setting up a new sample. I followed the pattern and added that var to the cleanup list.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -437,6 +437,13 @@ public static class ConfigurationKeys
...
+        /// </summary>
+        /// <seealso cref="TracerSettings.TagMongoCommands"/>
+        public const string TagMongoCommands = "SIGNALFX_INSTRUMENTATION_MONGODB_TAG_COMMANDS";
```
pellared, line 446: Can you update `docs/USAGE.md`? Also why `_COMMANDS` suffix and not e.g. `_DB_STATEMENT` ?

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
RassK: Added.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
```diff
@@ -366,6 +369,12 @@ internal static Scope CreateScope(object wireProtocol, object connection)
             {
                 if (wireProtocol.TryGetFieldValue("_command", out object command) && command != null)
                 {
+                    if (tracer.Settings.TagMongoCommands)
+                    {
+                        string document = command.ToString();
```
pellared: we are duplicating `mongodb.query` and `db.statement` tags
maybe we can assign the `db.statement` just below the  `mongodb.query`  assignement?

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
```diff
@@ -366,6 +369,12 @@ internal static Scope CreateScope(object wireProtocol, object connection)
...
+                        statement = document.Length > 1024 ? document.Substring(0, 1024) : document;
```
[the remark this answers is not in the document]
RassK: Removed truncation.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
RassK: Seems command is more towards the code where the value originates from... the statement comes from just reusing general db tags.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
RassK: um, seems I have somehow passed the way query is fetched in the main branch.

pellared:
Can you give hyperlinks to traces in Splunk o11y Cloud?

pellared (approved):
LGTM however it would be good to see how it looks in o11y Cloud to check if it draws the MongoDB on the "service map"

dszmigielski (approved):

RassK:
> LGTM however it would be good to see how it looks in o11y Cloud to check if it draws the MongoDB on the "service map"

I'll post it here as well.
https://app.signalfx.com/#/apm/traces/10f023939b10fb4c1e1ec70bb537248c

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
```diff
@@ -388,18 +422,19 @@ internal static Scope CreateScope(object wireProtocol, object connection)
...
                 var tags = new MongoDbTags();
-                scope = tracer.StartActiveWithTags(OperationName, serviceName: serviceName, tags: tags);
+                scope = tracer.StartActiveWithTags(operationName ?? DefaultOperationName, tags: tags);
                 var span = scope.Span;
+                span.LogicScope = DefaultOperationName;
                 span.Type = SpanTypes.MongoDb;
                 span.ResourceName = resourceName;
+                tags.DbType = SpanTypes.MongoDb;
                 tags.DbName = databaseName;
+                tags.DbStatement = statement;
```
lachmatt, line 437: I think this tag should be checked for in GetActiveMongoDbScope

RassK: Good notice!

@pellared @pjanotti currently in the main we filter out exact spans if there happens to be a nesting one. Upstream is filtering out all nested MongoDB tags, which is a way more performant due `GetTag` implementation. Wondering if we should actually do the same?

pellared: As I do not think it is critical right now I suggest creating a JIRA story and discuss it during a meeting during which @pjanotti will be present.

PR #191, opened 2021-11-02
ASP.NET semantic conventions adjustments

lachmatt:

- Adjusted ASPNET integrations to http semantic convention
- Fixed possible exception in W3CSpanContextPropagator
- Adjusted verification files after ADONET changes

@signalfx/instrumentation


tracer/src/Datadog.Trace/ClrProfiler/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -117,12 +117,6 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
                         resourceName = $"{httpMethod} /{routeUrl.ToLowerInvariant()}";
                     }
 
-                    if (string.IsNullOrEmpty(resourceName) && httpContext.Request.Url != null)
-                    {
-                        var cleanUri = UriHelpers.GetCleanUriPath(httpContext.Request.Url);
-                        resourceName = $"{httpMethod} {cleanUri.ToLowerInvariant()}";
-                    }
```
lachmatt, line 124: This could result in high cardinality in resourceName which I use as operationName for span, so after discussing it with Robert I decided to remove it

pellared:
Hyperlink for code review: https://github.com/signalfx/signalfx-dotnet-tracing/pull/191/files/767df836a7645ef1981fddd97ea598bd2d8e29bd 😄

tracer/src/Datadog.Trace/ClrProfiler/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -74,7 +74,7 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
                     var newResourceNamesEnabled = Tracer.Instance.Settings.RouteTemplateResourceNamesEnabled;
                     string host = httpContext.Request.Headers.Get("Host");
                     string httpMethod = httpContext.Request.HttpMethod.ToUpperInvariant();
-                    string url = httpContext.Request.RawUrl.ToLowerInvariant();
+                    string url = httpContext.Request.Url?.ToString(); // Upstream uses RawUrl, ie. the part of the URL following the domain information.
```
pellared, line 77: can the `Url` be null? I am curious why `Url?.`

tracer/src/Datadog.Trace/ClrProfiler/Integrations/AspNet/AspNetMvcIntegration.cs
pellared: However, I see that `RouteTemplateResourceNamesEnabled` is `false` by default for an unknown reason and the env var is not even documented. So I think that `resourceName = $"{httpMethod} /{routeUrl.ToLowerInvariant()}";` will not be executed.

pellared: Per discussion, we agree that we can be `true` by default and describe the env var in `docs/USAGE.md`. And @pjanotti can double-check everything later (after it is even merged).

lachmatt: will do in a separate PR

pellared (approved):

tracer/src/Datadog.Trace/ClrProfiler/Integrations/AspNet/AspNetMvcIntegration.cs
pellared: https://github.com/signalfx/signalfx-dotnet-tracing/pull/196

pjanotti: notice that upstream is also changing the default value for the env var

PR #192, opened 2021-11-03
Add option to disable CancelKeyPress event subscription

pjanotti:
The main purpose of this change is to test a case in which there is an exception when the finalizer of the handler is called. It seems that this is not actually the cause of the crash but a side effect. Building a version in which we can disable the subscription to CancelKeyPress so we can investigate further.

At this point it is not clear if this should be merged or not.


RassK:
BTW, in next-ver it's wrapped like this > 

```
try
{
    // Registering for the cancel key press event requires the System.Security.Permissions.UIPermission
    Console.CancelKeyPress += Console_CancelKeyPress;
}
catch (Exception ex)
{
    Log.Warning(ex, "Unable to register a callback to the Console.CancelKeyPress event.");
}
```

pjanotti:
This proved useful in general. Let's prepare a release with it.

RassK (approved):
LGTM

pellared (approved):

PR #196, opened 2021-11-04
Default RouteTemplateResourceNamesEnabled to true

pellared:
## What

`RouteTemplateResourceNamesEnabled` is `true` by default.

## Why

https://github.com/signalfx/signalfx-dotnet-tracing/pull/191#discussion_r742102774

Why we are considering changing it

1. It is what is currently done on `main`: https://github.com/signalfx/signalfx-dotnet-tracing/blob/af2d083175b6433d8ddce700c70a22c1817b7af7/src/Datadog.Trace/Configuration/TracerSettings.cs#L183 . In this PR: https://github.com/signalfx/signalfx-dotnet-tracing/pull/118

Why it may be good to leave as it is:

1. Upstream still defaults to `false`.
2. Personally I am not sure what is the advantage when comparing the `.NoFF.*.verified.txt` files with `.WithFF.*.verified.txt` files. @RassK can you find some examples there?


pellared:
I am postponing the changes in the golden files until we know if it is a desired change.

pjanotti (approved):

pellared:
It is already changed in upstream (PRs: 1961, 2012) so maybe it would be better to just "take it" when doing upstream sync?

pjanotti:
Agreed @pellared

PR #197, opened 2021-11-04
Remove Jaeger exporter

pellared:
## What

- Remove Jaeger exporter.
- Fix the missing `SIGNALFX_EXPORTER` row in the env var table in `USAGE.md`

## Why

Per https://github.com/signalfx/signalfx-dotnet-tracing/pull/185#discussion_r737340980

1. We do not want to rely on UDP. Zipkin exporter is safer
2. Quicker build time
3. Lower maintenance cost


dszmigielski (approved):
image
Love it :D

PR #198, opened 2021-11-05
Upstream sync 8920a1d..86c0bc5

pellared:
```
git cherry-pick 8920a1d..86c0bc5
```

git log:

```
commit 86c0bc5542afa42600c39931f741f3b1f7b4ddf2
Author: Lucas Pimentel-Ordyna <lucas.pimentel@datadoghq.com>
Date:   Thu Nov 4 14:04:54 2021 -0400

    Use UserKeep(2)/UserReject(-1) in the rules-based sampler (#1937)

commit ce1ba6dd716e325e96cda5769ce3036b1723a416
Author: Zach Montoya <zach.montoya@datadoghq.com>
Date:   Wed Nov 3 15:27:33 2021 -0700

    Add AssemblyLoadContext to logs for netcoreapp assembly builds (#1977)

commit a7c4b054afbd50348346dc2bcd4da42121458b87
Author: Zach Montoya <zach.montoya@datadoghq.com>
Date:   Wed Nov 3 10:02:32 2021 -0700

    Additional updates for the joint MSI (#1915)
    
    - Renames the installed product from ".NET Tracer" to ".NET Monitoring"
    - Renames the repo source directory and the installed on-disk directory from "tracer" to "Tracer"
    - Add `Activate-Installed_DotNet-Monitoring.bat` to the ContinuousProfiler folder of the installation
    - GUID updates:
      - Reset the beta MSI to use the same profiler GUID as the release MSI today
      - Native Loader (new loading mechanism) now only responds to the original Tracer GUID {846F5F1C-F9AE-4B07-969E-05C26BC060D8}
      - Native Tracer (old loading mechanism) responds to both old Tracer GUID and new GUID {50DA5EED-F1ED-B00B-1055-5AFE55A1ADE5} which will be called by the new Native Loader

commit 6cfaa55336f52ef09afe87bc093dd207738d9dab
Author: Jean-Baptiste Aviat <jb@datadoghq.com>
Date:   Tue Nov 2 23:37:15 2021 +0100

    Make vuln app database samples path work cross OS (#1824)
    
    Co-authored-by: Jean-Baptiste Aviat <jb@trin.local>

commit 016d0ed04aa13a4562b0a4652f40270b1e4538c5
Author: Zach Montoya <zach.montoya@datadoghq.com>
Date:   Mon Nov 1 11:38:51 2021 -0700

    Escape http.url tag in AspNetCore spans (#1938)
    
    When extracting the URL from an HttpRequest, make sure it is an escaped URL. This makes sure that reported URL's match between HTTP spans and AspNetCore spans.

commit c87da06cf691ea25bcb095d9f998a1796bd549c1
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Mon Nov 1 10:13:44 2021 +0000

    Treat release/ branches as the "main" branch (#1963)
    
    This ensures we run the full test suite etc on merges to release/2.0 and release/1.x

commit 18f6840145c09134c70baec2939a300694c6e684
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Fri Oct 29 13:36:01 2021 +0100

    "Update test package versions" PR has to run UpdateIntegrationsJson (#1954)
    
    * The "update test package versions" PR has to update the dependabot file too
    
    * Always raise the PR against master

commit 3b5b5dcc66d28dfac974f44faa60ba12cc37405e
Author: Kevin Gosse <krix33@gmail.com>
Date:   Fri Oct 29 12:36:33 2021 +0200

    Properly set sampling priority when using GetRequestStream (#1947)

    * Disable buffering for GetRequestStream test
    * Fix sampling priority
    * Add unit test to validate the trace id behavior

commit d0e42ae210ba2b58456a604ad2568310f37182f0
Author: Charles de Beauchesne <charles.de.beauchesne@gmail.com>
Date:   Fri Oct 29 11:36:44 2021 +0200

    Add system tests into CI (#1942)

commit 39baaf6259e7fc1398aec8de44dddce46f7953cc
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Thu Oct 28 15:04:26 2021 +0100

    Fix GitHub Actions workflows (#1946)

    * Try fix github workflow when dependabot creates PRs

    Also allow manual dispatch so we can update whenever we like

    * Auto tag the version bump commit on `release/*` branches too

    * Apply suggestions from code review

    Co-authored-by: Zach Montoya <zach.montoya@datadoghq.com>

    * Ensure we can run the workflow when doing manual bump

    Co-authored-by: Zach Montoya <zach.montoya@datadoghq.com>

commit 7e34ee7d766b254e3b5a08ae8123c6ee3df657e1
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Thu Oct 28 14:17:06 2021 +0100

    Fix condition in Datadog.Trace.proj breaking the build on master (#1950)

commit a31b145a3d4dbb6c2a7725e1f57b2ada34396468
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Thu Oct 28 10:41:39 2021 +0100

    Add a "LatestMajors" option for testing all major package versions (#1939)

    It's an in-between step between TestAllMinorPackageVersions, and only testing the default samples, and is now the default for the multi-api package tests

    In many cases, the results are the same as using the default samples, but it should give more test coverage in PRs generally

commit 30f5b071c5c773eb61beae2675051c229d0ed2a7
Author: macrogreg <macrogreg@users.noreply.github.com>
Date:   Wed Oct 27 11:29:59 2021 -0700

    Fix typo in comments (#1928)

    Just merged https://github.com/DataDog/dd-trace-dotnet/commit/ea5e046488c612f8d4a4bee04a7c66532efa58b3, and subsequently discovered a typo in the comments.
    Sorry about that.
    This fixes the typo.

commit 96f298cadc8cb3d26188d490e2350b622e7234b0
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Wed Oct 27 14:17:56 2021 +0100

    Refactor reverse duck-typing and add negative tests (#1900)

    * Add tests that we _can't_ duck type invalidly defined types

    We have lots of tests that we _can_ duck type certain things, but there are many cases where we should throw when attempting to Duck Cast. If we don't, we risk accidentally thinking we've duck cast correctly when we haven't.

    Obviously an even better approach would be to have an analyzer that detects this too, but that's a _lot_ more work

    * Add checks for method resolution that Duck attribute works with generic parameter names

    * Add some error tests for reverse duck typing

    * Add failing test for reverse proxy where virtual methods already have a method body

    * Add test that reverse-proxy properties work correctly (they don't currently)

    * Fix uninitialized fields and properties in reverse proxy

    Call base constructor if the proxy type is inheriting from a class or abstract class.

    * Change type returned by DuckType.Create() to an object

    If the ProxyType is a struct, then even if the implementation is correct, this will throw for struct proxies when it attempts to cast to an IDuckType (as the struct doesn't implement IDuckType). The cast is largely unnecessary anyway, 
and can be cast explicitly by the caller if necessary (and safe)

    * Fix cases in field and property duck typing that were being too lenient

    They were being lenient to handle _reverse_ duck typing scenarios, but we'll split those out for clarity and increase the strictness here

    Also add some warnings where we don't (can't) know if a conversion is valid, so it would result in a runtime error accessing the member

    * Add additional guard clauses for method duck typing

    - Catch when an argument changes from generic to non-generic in a generic method
    - Catch when the original method returns void, but the new method returns non-void

    * Skip tests for which we can't (or won't) detect failures

    (Primarily around return types and conversions)

    * Split reverse-duck-typing from forward-duck-typing

    While these cases are quite similar, there are a bunch of requirements and edge cases that make interleaving the implementation problematic. Some of the additional tests added highlighted this problem.

    To try and make changes to the duck-typing code safer, this separates both the implementation and the API of reverse duck typing from "forward" duck typing. This also fixes some of the implementation issues that were previously causing failures

    * Make [DuckReverseMethod] a mirror of [Duck] attribute

    - Extracted the common features to a DuckAttributeBase
    - Allow [DuckReverseMethod] to be used on properties
    - Update method resolution to use the new attributes
    - Fix method resolution for forward duck-typing when using Duck attribute and have generic parameters
    - Fix method resolution for reverse duck-typing
    - [DuckIgnore] is no longer used by reverse duck typing, as you must decorate all your reverse methods/properties anyway

    * Fix [DuckReverseMethod] usages

    * Update SelectTargetMethod implementation

    - Fix case where the duck type attribute parameters don't match method parameters count
    - Fix duck chaining from reverse duck types - need to do _reverse_ duck chaining here, not normal forward duck chaining
    - Fix reverse duck typing not finding
    - Fix invalid test (caught with above changes)

    * Handle additional incorrect usage case and add additional exception tests

    * Ensure IntegrationOptions catches DuckType exceptions when using DuckCast(Type), as well as DuckCast<T>

    * Refactor DuckType.Methods.cs to reduce duplication

    * Update Public API tests

    This added some public DuckTyping methods. They could be made internal, but their equivalents are already public, so I think it makes more sense to keep them all public for now

    * Update ExceptionsTests

    Introduce ReverseProxyMustImplementGenericMethodAsGenericException to make it easier to debug the issue
    Remove unused exception
    Add missing tests for ReverseProxyMissingMethodImplementationException
    Add some extra tests for code coverage purposes

    * Update incorrect comments

    * Fix public API

    * Comments from code review

    * Update reverse proxy properties implementation

    Reverse proxies for methods was previously broken.

    As for methods, when we're doing reverse proxying, we need to do the opposite ducktype/duckcast compared to forward proxying

    This also removes the DuckKind.Field for reverse proxies (as that doesn't make sense)
    Also, updated the "duck cast/duck chain" delegates to return the actual type that you end up with, for use by the caller.

    * Handle case where [DuckReverseMethod] attribute has different number of parameters than method

    This always points to a configuration error for reverse duck typing, whereas for forward duck typing this isn't _necessarily_ an error (as the target may have optional parameters, for example)

    Add an explicit check for the reverse duck type case to catch this.

    * Fix incorrect arguments in test

    * Use public types so we can test net452 in reverse proxy error tests

    * Fix incorrectly ignored `WrongArgumentTypeInterfaceImplementations.DuckChainReturnMethod` test

    * Fix Public API tests

    * Ensure we catch _all_ exceptions when creating a duck type

    Co-authored-by: Tony Redondo <tony.redondo@datadoghq.com>

commit cfd66df65ed9e7142b50673e8a163ae9e105cf4f
Author: Andrew Lock <andrew.lock@datadoghq.com>
Date:   Wed Oct 27 09:00:45 2021 +0100

    Small build fixes (#1936)

    * Fix build_in_docker.sh

    This was broken when we moved things into the tracer folder and wasn't fixed correctly

    * Always build AspNetCoreRazorPages when building single sample

    If you try and build a single sample on linux using (for example)

    ```
    ./tracer/build_in_docker.sh Clean BuildTracerHome BuildAndRunLinuxIntegrationTests -SampleName "LogsInjection.NLog"
    ```

    then the build will fail. That's because the linux integration tests depend on `Samples.AspNetCoreRazorPages`, but when a single sample name is provided, it won't be built.

    This change ensures we _always_ build `Samples.AspNetCoreRazorPages`, even when a specific sample is specified using `-SampleName <name>`

commit 442d28020c0ce5c2f84747c2451182fed0095f4f
Author: Kevin Gosse <krix33@gmail.com>
Date:   Tue Oct 26 17:38:30 2021 +0200

    Give some time for the aspnetcore process to exit in AppSec tests (#1926)

    * Give some time for the aspnetcore process to exit
    * Let AspNetBase dispose the agent (The agent must be disabled after the target process shuts down)

commit e3b32c7fc20f959cd691b0507125ecacbdd58eab
Author: Tony Redondo <tony.redondo@datadoghq.com>
Date:   Mon Oct 25 18:34:55 2021 +0200

    Refactor ModuleMetadata handling to reduce memory allocations (#1931)

    * Refactor ModuleMetadata creation from all loaded modules to only the required ones.

    * Adds std::future to the ModuleLoadFinished EnqueueProcessModule call.

    * Process module for ReJIT in the same thread as ModuleLoadFinished

    * Protect for shutdown deadlock.

    * Add support for retrieving memory dumps on integration tests hangs

    * Avoid calling RequestReJIT in the same thread of ModuleLoadFinished.

    * remove unused local variable

    * Change the memdump algorithm and disable it by default.

    * Ensure the promise is resolved on shutdown.
```


tracer/src/Datadog.Trace.ClrProfiler.Native/dllmain.cpp
```diff
@@ -23,7 +23,10 @@ extern "C"
         // {B4C89B0F-9908-4F73-9F59-0D77C5A06874}
         const GUID CLSID_CorProfiler = {0xb4c89b0f, 0x9908, 0x4f73, {0x9f, 0x59, 0xd, 0x77, 0xc5, 0xa0, 0x68, 0x74}};
 
-        if (ppv == NULL || rclsid != CLSID_CorProfiler)
+        // {50DA5EED-F1ED-B00B-1055-5AFE55A1ADE5}
+        const GUID CLSID_New_CorProfiler = {0x50da5eed, 0xf1ed, 0xb00b, {0x10, 0x55, 0x5a, 0xfe, 0x55, 0xa1, 0xad, 0xe5}};
```
pellared, line 27: New CLSID added. If we want to have a different value than upstream then I would prefer to do it in a separate PR.

tracer/integrations.json
```diff
@@ -2843,7 +2843,7 @@
           "maximum_patch": 65535
         },
         "wrapper": {
-          "assembly": "Datadog.Trace, Version=1.29.0.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
+          "assembly": "SignalFx.Tracing, Version=0.0.1.0, Culture=neutral, PublicKeyToken=e43a27c2023d388a",
```
RassK, line 2846: Should the version here be different? we used 0.0.1.0 in OTel, but SFX follows different path.

pellared: It should but we need a different Jira story for it. @pjanotti is it under some existing story?

RassK: nvm I though that in next-ver others are different, it's 0.0.1.0 in all places. So definitely another (new task) scope.

RassK (approved):
LGTM

dszmigielski (approved):

.github/workflows/auto_bump_test_package_versions.yml
```diff
@@ -33,6 +36,7 @@ jobs:
           branch: "bot/test-package-versions-bump"
           commit-message: "[Test Package Versions Bump]"
           delete-branch: true
+          base: master
           title: "[Test Package Versions Bump] Updating package versions "
           milestone: "${{steps.rename.outputs.milestone}}"
           reviewers: "DataDog/apm-dotnet"
```
pjanotti, line 42: We can change the reviewers here - although we are not using this yet.

pjanotti (approved):
LGTM

tracer/integrations.json
pjanotti: I tried updating this via nuke and it works as expected, updating the version should be a small story.

.github/workflows/auto_bump_test_package_versions.yml
RassK: seeing this comment here, some rows up `base: master`, should it be now `base: main`?

pjanotti: yes, but not right now.

pjanotti (approved):

PR #199, opened 2021-11-05
Kafka semantic conventions

dszmigielski:
Added messaging.system and messaging.destination tags to Kafka integration.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
@@ -126,7 +126,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
 
                 var span = scope.Span;
                 span.LogicScope = KafkaConstants.ConsumeOperationName;
-                span.LogicScope = KafkaConstants.ProduceOperationName;
                 tags.Destination = topic;
                 tags.System = SystemName;
                 span.Type = SpanTypes.Queue;
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="KafkaHelper.cs" company="Datadog">
+// <copyright file="KafkaHelper.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
@@ -14,7 +14,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
 {
     internal static class KafkaHelper
     {
-        private const string SystemName = "rabbitmq";
+        private const string SystemName = "kafka";
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(KafkaHelper));
         private static bool _headersInjectionEnabled = true;
 
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
@@ -16,7 +16,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
     {
         private const string SystemName = "kafka";
         private const string OperationReceive = "receive";
-        private const string OperationSend = "send";
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(KafkaHelper));
         private static bool _headersInjectionEnabled = true;
 
@@ -54,7 +53,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
                 span.LogicScope = KafkaConstants.ProduceOperationName;
                 tags.Destination = topicPartition?.Topic;
                 tags.DestinationKind = SpanTypes.Queue;
-                tags.Operation = OperationSend;
                 tags.System = SystemName;
                 span.Type = SpanTypes.Queue;
                 span.ResourceName = resourceName;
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -12,6 +14,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
 {
     internal static class KafkaHelper
     {
+        private const string SystemName = "rabbitmq";
```
pjanotti, line 17: this should be kafka

tracer/src/Datadog.Trace/Tagging/KafkaTags.cs
```diff
@@ -3,12 +3,14 @@
...
 
+// Modified by Splunk Inc.
+
 using System;
 using Datadog.Trace.ExtensionMethods;
 
 namespace Datadog.Trace.Tagging
 {
-    internal class KafkaTags : InstrumentationTags
+    internal class KafkaTags : MessagingTags
```
pjanotti, line 13: This should be the case of the test failures: it changes the number of tags and at least one of the errors is `Mismatch between readonly properties and tags count for type Datadog.Trace.Tagging.KafkaTags`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -114,15 +118,16 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
                     }
                 }
 
-                string serviceName = tracer.Settings.GetServiceName(tracer, KafkaConstants.ServiceName);
-
                 var tags = new KafkaTags(SpanKinds.Consumer);
 
-                scope = tracer.StartActiveWithTags(KafkaConstants.ConsumeOperationName, parent: propagatedContext, tags: tags, serviceName: serviceName);
+                scope = tracer.StartActiveWithTags(KafkaConstants.ConsumeOperationName, parent: propagatedContext, tags: tags);
```
RassK, line 123: fixed operation name will group every kafka message... although it's in main, it should be more towards ResourceName.

dszmigielski: well, I can replace it with resource name if that's what we want

dszmigielski: I will create new PR for that :)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -36,18 +39,19 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
...
-                    serviceName: serviceName,
                     finishOnClose: finishOnClose);
 
                 string resourceName = $"Produce Topic {(string.IsNullOrEmpty(topicPartition?.Topic) ? "kafka" : topicPartition?.Topic)}";
 
                 var span = scope.Span;
+                span.LogicScope = KafkaConstants.ProduceOperationName;
+                tags.Destination = topicPartition?.Topic;
+                tags.System = SystemName;
                 span.Type = SpanTypes.Queue;
```
RassK, line 55: Seems like a bug in main, I think it should be topic.

RassK: > Seems like a bug in main, I think it should be topic.

`tags.DestinationKind` can be easily added. Should be also `"topic"`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -114,15 +118,16 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
...
 
-                scope = tracer.StartActiveWithTags(KafkaConstants.ConsumeOperationName, parent: propagatedContext, tags: tags, serviceName: serviceName);
+                scope = tracer.StartActiveWithTags(KafkaConstants.ConsumeOperationName, parent: propagatedContext, tags: tags);
 
                 string resourceName = $"Consume Topic {(string.IsNullOrEmpty(topic) ? "kafka" : topic)}";
 
                 var span = scope.Span;
+                span.LogicScope = KafkaConstants.ConsumeOperationName;
+                tags.Destination = topic;
+                tags.System = SystemName;
```
RassK, line 130: `tags.Operation = "receive"` could be easily added here (only for consumption) as well.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -36,18 +41,21 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
...
-                    serviceName: serviceName,
                     finishOnClose: finishOnClose);
 
                 string resourceName = $"Produce Topic {(string.IsNullOrEmpty(topicPartition?.Topic) ? "kafka" : topicPartition?.Topic)}";
 
                 var span = scope.Span;
+                span.LogicScope = KafkaConstants.ProduceOperationName;
+                tags.Destination = topicPartition?.Topic;
+                tags.DestinationKind = SpanTypes.Queue;
+                tags.Operation = OperationSend;
```
RassK, line 57: > A string identifying the kind of message consumption as defined in the Operation names section above. If the operation is "send", this attribute MUST NOT be set, since the operation can be inferred from the span kind in that case.

I followed this instruction when doing RabbitMQ.

RassK (approved):
LGTM

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -114,16 +120,19 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
                     }
                 }
 
-                string serviceName = tracer.Settings.GetServiceName(tracer, KafkaConstants.ServiceName);
```
RassK, line 117: as agreed in the meeting this needs to be reverted.

PR #200, opened 2021-11-10
Set db.statement in ADO.NET integrations

lachmatt:
Set db.statement for ado.net-based integrations to command's text

Samples:
sqlserver: https://app.signalfx.com/#/apm/traces/16557cd0ec82d95e11f5141ce651f911
postgresql: https://app.signalfx.com/#/apm/traces/2362c60b654f98f90d189eed852c2d5e

@signalfx/instrumentation


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
```diff
@@ -75,6 +77,8 @@ public void SubmitsTracesWithNetStandard()
                     Assert.Equal(SpanTypes.Sql, span.Type);
                     Assert.Equal(dbType, span.Tags?[Tags.DbType]);
                     Assert.False(span.Tags?.ContainsKey(Tags.Version), "External service span should not have service version tag.");
+                    Assert.NotNull(span.Tags?[Tags.DbStatement]);
+                    Assert.NotEmpty(span.Tags[Tags.DbStatement]);
```
RassK, line 81: here `SpanExpectations.GetTag()` could be used. Although I'm wondering the null check for `span.Tags`, Tags dictionary should be always present... or `SpanExpectations.GetTag()` can be modified to account this.

tracer/src/Datadog.Trace/ClrProfiler/ScopeDBFactory.cs
```diff
@@ -62,10 +64,7 @@ public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
             {
                 Span parent = tracer.ActiveScope?.Span;
 
-                if (parent != null &&
-                    parent.Type == SpanTypes.Sql &&
-                    parent.GetTag(Tags.DbType) == _dbTypeName &&
-                    parent.ResourceName == command.CommandText)
+                if (ScopeFactory.IsAlreadyInstrumented(parent, _dbTypeName, command.CommandText))
```
pjanotti, line 67: I like the refactor just need to keep an eye for changes on upstream on the criteria happening on upstream.

pjanotti: Also it makes more sense to either have the method on `ScopeDBFactory` or to change its name to saying that indicates that it is checking for DB related settings.

tracer/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -44,9 +46,15 @@ public static void SetTraceSamplingPriority(this Span span, SamplingPriority sam
         /// <param name="command">The db command to get tags values from.</param>
         public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
         {
-            span.ResourceName = command.CommandText;
+            var commandText = command.CommandText;
+            span.ResourceName = commandText;
             span.Type = SpanTypes.Sql;
 
+            if (!string.IsNullOrEmpty(commandText))
```
pjanotti, line 53: We don't export null tags so it doesn't seem necessary to have this check here.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
pjanotti: I don't see a case that we have a span with 0 tags. Upstream may supports that but all of our instrumentation will add at least one tag.

tracer/src/Datadog.Trace/ClrProfiler/ScopeDBFactory.cs
lachmatt: I placed it here, because there is already a method shared between `ScopeFactory` and `ScopeDbFactory`. I will change the name to indicate its db-related.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
```diff
@@ -43,7 +44,7 @@ public void SubmitsTraces()
                     Assert.Equal(expectedServiceName, span.Service);
                     Assert.Equal(SpanTypes.Sql, span.Type);
                     Assert.Equal(dbType, span.Tags?[Tags.DbType]);
-                    Assert.False(span.Tags?.ContainsKey(Tags.Version), "External service span should not have service version tag.");
```
pellared, line 46: why is this assertion removed?

lachmatt: It was removed with changes to setting `serviceName`, which I reverted as @Kielek is addressing it in his PR.

pellared: then revert it please and let's have a dedicated PR to address it 👍

pellared (approved):

tracer/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -44,9 +46,12 @@ public static void SetTraceSamplingPriority(this Span span, SamplingPriority sam
         /// <param name="command">The db command to get tags values from.</param>
         public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
         {
-            span.ResourceName = command.CommandText;
+            var commandText = command.CommandText;
+            span.ResourceName = commandText;
             span.Type = SpanTypes.Sql;
 
+            span.SetTag(Tags.DbStatement, commandText);
```
RassK, line 53: I wouldn't use this `.SetTag()` if possible. Seems unnecessary and expensive way to locate setter. 
If possible, check if tags returned below is of certain type that contains DbStatement property. Then the assignment would be very fast.

RassK: Agreed to resolve later, as it needs more refactoring due `GetTagsFromDbCommand` doesn't return appropriate type.

RassK (approved):
LGTM

tracer/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
pjanotti: Can change the signature and pass the tags too?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
```diff
@@ -44,6 +45,7 @@ public void SubmitsTraces()
                     Assert.Equal(SpanTypes.Sql, span.Type);
                     Assert.Equal(dbType, span.Tags?[Tags.DbType]);
                     Assert.False(span.Tags?.ContainsKey(Tags.Version), "External service span should not have service version tag.");
+                    Assert.Contains(Tags.DbStatement, (IDictionary<string, string>)span.Tags);
```
pjanotti, line 48: Assert.True with ContainsKey like previous line seems more readable to me - in this case without the assert message.

pjanotti (approved):
LGTM, two suggestions one for the tests other to avoid the call mentioned by Rasmus (perhaps in a separate PR).

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
pellared: Without the custom assertion message, the test failure message is really awkward.
I suggest using `Assert.Contains` or `Assert.True/False` with an assert message.

lachmatt: I used `Assert.Contains` on purpose, as there is more context in case of failure, compared to `Assert.True`, unless we always add assertion message to the latter.

PR #203, opened 2021-11-14
Automate release for next-ver branch

pjanotti:
Automate the generation of releases for the next-ver branch. It includes the Alpine package. 

It doesn't include automatic generation of release notes and version update. Upstream has some related actions for that but keeping it very simple here so we can start publish releases for earlier test of next-ver. All releases are created via this, are marked as pre-release and also as a draft so final publication requires a manual approval.

When next-ver becomes the default branch we can improve the process by triggering this type of action manually.

- Run of publish release tags: https://github.com/pjanotti/signalfx-dotnet-tracing/actions/runs/1459480288
- Example of release: https://github.com/pjanotti/signalfx-dotnet-tracing/releases/tag/untagged-c2b1812b905167a33226

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -73,7 +73,7 @@ jobs:
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\/v}
       # TODO: Extract release notes
       - name: Create Release
-        run: gh release create v${{ steps.get_version.outputs.VERSION }} --draft artifacts/signalfx* artifacts/x64/en-us/*.msi  artifacts/x86/en-us/*.msi
+        run: gh release create v${{ steps.get_version.outputs.VERSION }} --draft --prerelease artifacts/signalfx* artifacts/x64/en-us/*.msi  artifacts/x86/en-us/*.msi
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -48,7 +48,6 @@ jobs:
       - run: git config --system core.longpaths true
       - uses: actions/checkout@v2
       - uses: microsoft/setup-msbuild@v1.0.2
-      - name: Run 'Clean BuildTracerHome ZipTracerHome'
         run: tracer\build.cmd Clean BuildTracerHome PackageTracerHome
         shell: cmd
       - uses: actions/upload-artifact@v2
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -2,13 +2,13 @@ name: Publish Release Artifacts
 
 on:
   push:
-    branches:
-    - next-ver
     tags:
     - 'v*'
 
 jobs:
   linux_build:
+    name: Linux Build
+    if: github.event.base_ref == 'refs/heads/next-ver'
     runs-on: ubuntu-20.04
     timeout-minutes: 30
     strategy:
@@ -44,6 +44,7 @@ jobs:
 
   windows_build:
     name: Windows Build
+    if: github.event.base_ref == 'refs/heads/next-ver'
     runs-on: windows-2019
     steps:
       - run: git config --system core.longpaths true
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -3,7 +3,7 @@ name: Publish Release Artifacts
 on:
   push:
     branches:
-    - next-ver-automate-releases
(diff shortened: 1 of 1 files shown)


Kielek:
@pjanotti, the build used in https://github.com/pjanotti/signalfx-dotnet-tracing/actions/runs/1459480288#artifacts contains installers and libraries which are not signed by any certificate.

I am not sure what is the practice related to this stuff, but we should consider it.

pellared:
> @pjanotti, the build used in https://github.com/pjanotti/signalfx-dotnet-tracing/actions/runs/1459480288#artifacts contains installers and libraries which are not signed by any certificate.
> 
> I am not sure what is the practice related to this stuff, but we should consider it.

FYI https://github.com/signalfx/gdi-specification/blob/main/specification/repository.md

> - MAY have a `RELEASING.md` file that documents the release process, but this
>   file MUST NOT document private processes. For repositories that use private release
>   jobs, the `RELEASING.md` file SHOULD be absent or, if included, just contain the following note:
>   > The release process involves signing built packages and binaries and thus
>   > must be kept private and not exposed publicly.

we can discuss it during sync meeting

.github/workflows/artifacts.yml
```diff
@@ -1,10 +1,16 @@
-name: Build artifacts
+name: Publish Release Artifacts
+
 on:
-  workflow_dispatch:
+  push:
+    branches:
+    - next-ver
```
pellared, line 6: I do not think we want to make a release when pushing to next-ver branch

.github/workflows/artifacts.yml
```diff
@@ -70,12 +68,12 @@ jobs:
...
         id: get_version
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\/v}
-      - name: Extract release notes
-        run: awk '/## \[${{ steps.get_version.outputs.VERSION }}\]/,0' docs/CHANGELOG.md >> notes.md
-      - run: gh release create v${{ steps.get_version.outputs.VERSION }} -F notes.md artifacts/otel* artifacts/x64/en-us/opentelemetry* artifacts/x86/en-us/opentelemetry*
+      # TODO: Extract release notes
+      - name: Create Release
+        run: gh release create v${{ steps.get_version.outputs.VERSION }} --draft --prerelease artifacts/signalfx* artifacts/x64/en-us/*.msi  artifacts/x86/en-us/*.msi
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
pellared, line 78: I think we can update the `ci` workflow and use https://stackoverflow.com/a/58478262/859275 to conditionally create a GH release only when a tag is pushed.

And this workflow could be removed. The `ci` workflow already creates the needed artifacts if I remember correctly.

pjanotti:
@Kielek 
> @pjanotti, the build used in https://github.com/pjanotti/signalfx-dotnet-tracing/actions/runs/1459480288#artifacts contains installers and libraries which are not signed by any certificate.

We haven't added a certificate for the releases but it is a good thing to the backlog. 

@RassK could you please create a story for that.

.github/workflows/artifacts.yml
pjanotti: Removed the leftover. Adding conditional to ensure that this only runs when a tag is pushed to `next-ver`, for now we don't want this to run for main.

.github/workflows/artifacts.yml
pjanotti: Added a separate one to add Alpine that is not in ci. Since this runs only when a tag is pushed I'm not that worried about the cost of a separate one (only maintenance).

pellared: Can you test the workflow on your fork? I am not sure that the condition would work.

.github/workflows/artifacts.yml
```diff
@@ -46,13 +48,8 @@ jobs:
...
       - uses: actions/checkout@v2
       - uses: microsoft/setup-msbuild@v1.0.2
-      - run: dotnet --list-sdks
-        shell: cmd
-      - run: choco install wixtoolset
-        shell: cmd
-      - run: tracer\build.cmd BuildTracerHome
-        shell: cmd
-      - run: tracer\build.cmd PackageTracerHome
+      - name: Run 'Clean BuildTracerHome ZipTracerHome'
```
pellared, line 51: can we remove the `name` filed? 
1. it is misleading as `ZipTracerHome` is here instead of `PackageTracerHome`
2. it does not give any value

pellared (approved):

pellared (approved):

Kielek (approved):

PR #204, opened 2021-11-15
Adjust service name naming convention

Kielek:
Adjust service name naming convention for all cases
to avoid same changes in all instrumentations

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
@@ -422,12 +422,14 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                 Log.Warning(ex, "Unable to access IWireProtocol.Command properties.");
             }
 
+            string serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
+
             Scope scope = null;
 
             try
             {
                 var tags = new MongoDbTags();
-                scope = tracer.StartActiveWithTags(operationName ?? DefaultOperationName, tags: tags);
+                scope = tracer.StartActiveWithTags(operationName ?? DefaultOperationName, serviceName: serviceName, tags: tags);
                 var span = scope.Span;
                 span.LogicScope = DefaultOperationName;
                 span.Type = SpanTypes.MongoDb;
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
@@ -139,7 +139,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests.AWS
                     .ExcludingMissingMembers()
                     .ExcludingDefaultSpanProperties()
                     .AssertMetricsMatchExcludingKeys("_dd.tracer_kr", "_sampling_priority_v1")
-                    .AssertTagsMatchAndSpecifiedTagsPresent("deployment.environment", "aws.requestId", "aws.queue.url", "runtime-id"));
+                    .AssertTagsMatchAndSpecifiedTagsPresent("deployment.environment", "aws.requestId", "aws.queue.url"));
             }
         }
 
@@ -151,7 +151,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests.AWS
                 {
                     Name = "sqs.request",
                     Resource = $"SQS.{operationName}",
-                    Service = "Samples.AWS.SQS-aws-sqs",
+                    Service = "Samples.AWS.SQS",
                     Tags = new()
                     {
                         { "component", "aws-sdk" },
@@ -162,10 +162,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests.AWS
                         { "http.method", "POST" },
                         { "http.status_code", "200" },
                     },
-                    Metrics = new()
-                    {
-                        { "_dd.top_level", 1 }
-                    },
+                    Metrics = new(),
                     Type = SpanTypes.Http,
                 };
             }
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
@@ -28,7 +28,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests.AdoNet
             const int expectedSpanCount = 17;
             const string dbType = "postgresql";
             const string expectedOperationName = dbType + ".query";
-            const string expectedServiceName = "Samples.Dapper-" + dbType;
+            const string expectedServiceName = "Samples.Dapper";
 
             int agentPort = TcpPortProvider.GetOpenPort();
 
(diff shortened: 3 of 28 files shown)


tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
```diff
@@ -422,12 +422,14 @@ internal static Scope CreateScope(object wireProtocol, object connection)
...
 
+            string serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
+
             Scope scope = null;
 
             try
             {
                 var tags = new MongoDbTags();
-                scope = tracer.StartActiveWithTags(operationName ?? DefaultOperationName, tags: tags);
+                scope = tracer.StartActiveWithTags(operationName ?? DefaultOperationName, serviceName: serviceName, tags: tags);
```
pjanotti, line 432: Can we omit the serviceName parameter instead?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -139,7 +139,7 @@ public void SubmitsTraces(string packageVersion)
                     .ExcludingMissingMembers()
                     .ExcludingDefaultSpanProperties()
                     .AssertMetricsMatchExcludingKeys("_dd.tracer_kr", "_sampling_priority_v1")
-                    .AssertTagsMatchAndSpecifiedTagsPresent("deployment.environment", "aws.requestId", "aws.queue.url", "runtime-id"));
```
pjanotti, line 142: Why `runtime-id` disappears? It is a tag that at first doesn't seem related to the service name.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -162,10 +162,6 @@ public static AwsSqsSpan GetDefault(string operationName)
                         { "http.method", "POST" },
                         { "http.status_code", "200" },
                     },
-                    Metrics = new()
```
pjanotti, line 165: Since we plan to also use metrics in the future why is this being removed?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
```diff
@@ -43,7 +43,7 @@ public void SubmitsTraces()
                     Assert.Equal(expectedServiceName, span.Service);
                     Assert.Equal(SpanTypes.Sql, span.Type);
                     Assert.Equal(dbType, span.Tags?[Tags.DbType]);
-                    Assert.False(span.Tags?.ContainsKey(Tags.Version), "External service span should not have service version tag.");
+                    Assert.True(span.Tags?.ContainsKey(Tags.Version));
```
pjanotti, line 46: I don't understand what caused this change in behavior. Could you please clarify?

pjanotti (commented):
Excellent @Kielek - I know that is still a draft but here it goes some initial review.

pjanotti:
I actually expected a bunch of the tests to fail with diffs in the verify files but at quick glance they didn't. It seems that the few failures are related to AWS instrumentation.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
Kielek: It is added to the TopLevel spans only. Defined as span without Parent or span with Parent with different name.
 https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/src/Datadog.Trace/Span.cs#L131

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
Kielek: Sure, it is better described under #201 
```
Version tags - upstream introduces concept of version tag which was not available in main, for details check https://github.com/DataDog/dd-trace-dotnet/pull/716/files#diff-6508679470ef5919cdbaae772dab25f342463948fc3b0e495528e967508dc427R348-R351.
If we decided to change the name and make it DefaultServiceName then the version will be sent as a tag. The version is taken from SIGNALFX_VERSION (which is also new concept.
There is few possible solution
allow to accept version tags for Redis (by removing env. variable https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/test/test-applications/integrations/Samples.StackExchange.Redis/Properties/launchSettings.json#L16 or change in unit tests https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/StackExchangeRedisTests.cs#L281)
fully remove version tags from vnext
redefine logic here, how? - https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/src/Datadog.Trace/Tracer.cs#L484-L488
On main there is Tag version but defined as signalfx
```

There is separate story to handle the version tag correctly.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
Kielek: _dd.top_level is added only to top level spans. https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/src/Datadog.Trace/Tagging/TagsList.cs#L345-L350
Probably I have to bring back the Metrics property to pass these tests.

Kielek:
> I actually expected a bunch of the tests to fail with diffs in the verify files but at quick glance they didn't. It seems that the few failures are related to AWS instrumentation.

I have removed Metrics property from AwsSqsTests->Expected tagList. It leads to NullReferenceException. Fixed already.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
lachmatt: I think you need to do the same for #186

Kielek: @lachmatt, thanks, fixed

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/DapperTests.cs
lachmatt: I'd prefer to use `Assert.Contains` instead of `Assert.True` here, as described [here](https://github.com/signalfx/signalfx-dotnet-tracing/pull/200#discussion_r750141043)

Kielek: fixed, all places

tracer/src/Datadog.Trace/ClrProfiler/Integrations/MongoDbIntegration.cs
RassK: Why reverting, when it was initially agreed to omit the param instead?

Kielek: I have following reasons:
1. To be closer with the code to the upstrream
2. To have same behavior as in other instrumentations.
3. To bring back functionality related to custom mapping to GraphQ and MongoDB instrumentations.(SIGNALFX_TRACE_SERVICE_MAPPING),

pjanotti (approved):

PR #209, opened 2021-11-18
Set kafka operation name to resource name (containing topic)

dszmigielski:
Fixing kafka operation name as decided here:
https://github.com/signalfx/signalfx-dotnet-tracing/pull/199#discussion_r745884771

Samples:
Old: https://app.signalfx.com/#/apm/traces/00000000000002674811859898252523
New: https://app.signalfx.com/#/apm/traces/00000000000002537002085954705795


Kielek (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -43,14 +43,14 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
                 string serviceName = settings.GetServiceName(tracer, KafkaConstants.ServiceName);
                 var tags = new KafkaTags(SpanKinds.Producer);
 
+                string resourceName = $"Produce Topic {(string.IsNullOrEmpty(topicPartition?.Topic) ? "kafka" : topicPartition?.Topic)}";
+
                 scope = tracer.StartActiveWithTags(
-                    KafkaConstants.ProduceOperationName,
+                    resourceName,
```
RassK, line 49: I would use better `KafkaConstants.ProduceOperationName + topic` if exists. More similar to what's in other places.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -43,8 +43,9 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
                 string serviceName = settings.GetServiceName(tracer, KafkaConstants.ServiceName);
                 var tags = new KafkaTags(SpanKinds.Producer);
 
+                var operationName = string.IsNullOrEmpty(topicPartition?.Topic) ? KafkaConstants.ProduceOperationName : $"{KafkaConstants.ConsumeOperationName} {topicPartition.Topic}";
```
RassK, line 46: typo

```suggestion
                var operationName = string.IsNullOrEmpty(topicPartition?.Topic) ? KafkaConstants.ProduceOperationName : $"{KafkaConstants.ProduceOperationName} {topicPartition.Topic}";
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/KafkaTests.cs
```diff
@@ -60,11 +60,11 @@ public void SubmitsTraces(string packageVersion)
             // Due to manual/autocommit behaviour
             allSpans.Should().HaveCountGreaterOrEqualTo(TotalExpectedSpanCount);
 
-            var allProducerSpans = allSpans.Where(x => x.Name == "kafka.produce").ToList();
+            var allProducerSpans = allSpans.Where(x => x.LogicScope == "kafka.produce").ToList();
```
RassK, line 63: 1 notable thing maybe more. Currently we lose all span name checks here, which I think is important to have.
I think it's fine to get the spans by LogicScope, but just maybe add Name assertions.

dszmigielski: Good point

RassK (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -34,7 +34,7 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
 
                 var parent = tracer.ActiveScope?.Span;
                 if (parent is not null &&
-                    parent.OperationName == KafkaConstants.ProduceOperationName &&
+                    parent.LogicScope == KafkaConstants.ProduceOperationName &&
```
RassK, line 37: Seems last minute bug, didn't catch it before. When parent is a propagated span, I think there is no LogicScope?

dszmigielski: I reverted it back to use OperationName, but using "StartsWith".

pellared (approved):
LGTM

However, I would love to have the description updated to at least contain samples before and after the change (like here: https://github.com/signalfx/signalfx-dotnet-tracing/pull/211)

pellared (approved):
Something is strange with `Parent ID 0000000000000001` in our UI. Probably worth investigating as a separate issue.

PR #211, opened 2021-11-19
aerospike semantic convention

Kielek:
Aerospike semantic adjustment.
There is no aerospike instrumentation on main.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/database.yaml does not have exact specification for aerospike. Only `db.system` is defined.

Proposed changes from upstream
Change following tags
 * `aerospike.key` -> `db.aerospike.key`
 * `aerospike.namespace` -> `db.aerospike.namespace`
 * `aerospike.setname` -> `db.aerospike.setname`
 * `aerospike.userkey` -> `db.aerospike.userkey`
 * added `db.system` as `aerospike`

Operation name is always `aerospike.command`. Following tags are not added to spans `db.statement` and `db.command`. Aerospike internally contains data which could be used to populate this values but it is not available yet for current instrumentation implementation.

before changes: https://app.signalfx.com/#/apm/traces/44d62583834aa8290a4de2b7871189a8
after changes:  https://app.signalfx.com/#/apm/traces/743362a493e14dfd448d454259a156de

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
+// Modified by Splunk Inc.
+
 using System;
 using System.Text;
 using Datadog.Trace.Configuration;
@@ -38,6 +40,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Aerospike
                 var serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
                 scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName);
                 var span = scope.Span;
+                span.LogicScope = OperationName;
 
                 if (target.TryDuckCast<HasKey>(out var hasKey))
                 {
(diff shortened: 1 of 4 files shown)


lachmatt (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
```diff
@@ -38,6 +40,7 @@ public static Scope CreateScope<TTarget>(Tracer tracer, TTarget target)
                 var serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
                 scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName);
                 var span = scope.Span;
+                span.LogicScope = OperationName;
```
pellared, line 43: if we assign the `LogicScope` then we could assign some better operation (span) name.
based on https://docs.aerospike.com/docs/architecture/data-model.html I think that a name containing a `namespace` and `set` could be better

still - not for this PR and even not sure if worth implementing at all

pellared (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
Kielek: Agree, not this PR.
Aerospike allows to operate on multiple namespaces/sets in one operation.
Also internally in client there is available operation name like similar to READ/WRITE, etc.

PR #208, opened 2021-11-18
Upstream sync c30c3851..3f2036c

Kielek:
Upstream sync c30c3851..3f2036c

In scope of this task I reverted support for WCF instrumentation with System.ServiceModel versuin 5.0
This support was introduced in scope of https://github.com/signalfx/signalfx-dotnet-tracing/commit/215c5a4cd1bc8720749ac3a5fa6732d15cdea572 on main. Then in next-ver in scope of
https://github.com/signalfx/signalfx-dotnet-tracing/commit/713ab40039346fb4d4b5bf2223e4a0aa25f7628d

Our upstream is not supporting such version: https://github.com/DataDog/dd-trace-dotnet/commit/d009bf0e90018a34b5322a34abdba79228979cd9

What is more I was not able to find this version on my machine nor the Internet.

PR #212, opened 2021-11-22
Upstream doc update

Kielek:
Fix documentation for upstream sync.
Fix leftover from last upstream.

### DIFF
--- a/tracer/integrations.json
+++ b/tracer/integrations.json
@@ -2759,7 +2759,7 @@
           "minimum_major": 4,
           "minimum_minor": 0,
           "minimum_patch": 0,
-          "maximum_major": 5,
+          "maximum_major": 4,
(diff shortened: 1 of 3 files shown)


tracer/integrations.json
```diff
@@ -2759,7 +2759,7 @@
           "minimum_major": 4,
           "minimum_minor": 0,
           "minimum_patch": 0,
-          "maximum_major": 5,
+          "maximum_major": 4,
```
dszmigielski, line 2762: Why have we downgraded supported wcf version? :o

Kielek: See first comment in #208.
In my opinion, fixing old mistake.

dszmigielski (approved):

pellared (approved):

PR #214, opened 2021-11-22
SignalFx Tracing Library and Version

Kielek:
Introduce `signalfx.tracing.library` (also interpreted as a `otel.library.name`) and `signalfx.tracing.version` (`otel.library.version`) for root spans.

Difference against main branch
`TracerConstants` defines hardcoded version instead of taking it from assembly. There is a task `VerifyChangedFilesFromVersionBump` to verify update of this version when upgrading version. I would consider that it is enough.

Based on https://github.com/signalfx/signalfx-dotnet-tracing/pull/58/files

before changes: https://app.signalfx.com/#/apm/traces/4b0ee2cbdb556b361ef25375f3efbcad
after changes: https://app.signalfx.com/#/apm/traces/2c5be67f7b981939033d4d24c032b001


RassK (approved):
LGTM

PR #215, opened 2021-11-22
[WIP] Apply OTel convension for HTTP span error status

pellared:
## What

Mark HTTP client spans with `500-599` status codes as ERROR.

## Why

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status


pellared:
There is no existing automated test for this scenario. Do you think we should modify the `Samples.WebRequest` and `WebRequestTests`?

RassK (approved):

lachmatt (approved):

PR #221, opened 2021-11-23
Verification files update

lachmatt:
## Why

In order to have up-to-date verification files.

## What

Regenerated verification files for tests not being ran during CI.

## Tests
n/a


tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_branch_not-found_statusCode=404.verified.txt
```diff
@@ -2,7 +2,7 @@
   {
     TraceId: Id_1,
     SpanId: Id_2,
-    Name: GET /branch/not-found,
+    Name: HTTP GET,
```
RassK, line 5: Can this place be double checked, if it's the worst case scenario?

tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt
```diff
@@ -2,25 +2,27 @@
   {
     TraceId: Id_1,
     SpanId: Id_2,
-    Name: aspnet_core.request,
+    Name: GET status-code-string/{statusCode},
+    LogicScope: aspnet_core.request,
```
Comment (RassK) [inline tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt:6]: not sure if this needs to be here at all (LogicScope)

tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_branch_not-found_statusCode=404.verified.txt
RassK: Seems like something is wrong, as there is no way that MVC doesn't have a path.

RassK: Verified with @lachmatt 2 different things > 
1. Spans from selfhosted app without mvc / api layer
2. Azure Functions spans. 

In theory Azure functions seem to have routing config with variable placeholders, so if it's retrievable we might build something on top of it. (Will be a separate task)

pellared:
If we want to support .NET Core 2.1 and 3.0 then I suggest 

1. updating the CI to run the integration tests against these frameworks as well
2. update the "script" in `PULL_FROM_UPSTREAM.md` to include these frameworks

I would personally first want to know if we plan to officially support these frameworks. I think that @pjanotti was saying that we do not have to... but maybe it was only for the beginning? We should double-check with Ivo.

tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt
Comment (pellared) [reply tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt:6]: Personally, I think we can keep those. The `LogicScope` can help us identify the "origin" of the span in the codebase.

tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_branch_not-found_statusCode=404.verified.txt
lachmatt: I will create a separate task for it

tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt
Comment (lachmatt) [reply tracer/test/snapshots/AspNetCoreIisMvc21Tests.OutOfProcess.NoFF.__path=_status-code-string_[200]_statusCode=500.verified.txt:6]: Let's keep it as it is for now.

pellared (approved):

PR #222, opened 2021-11-24
B3 as deafult propagator

Kielek:
## Why

Have the same propagator as already released SignalFx library.

## What

Changing the default Propagator to `B3`

## Tests

1. Integration tests
2. Locally `dev/envvar.sh` does not set SIGNALFX_PROPAGATORS and it works 🎉


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -762,9 +762,7 @@ private static HashSet<string> GetPropagators(IConfigurationSource source)
 
             if (!propagators.Any())
             {
-                // TODO: Default to W3C (be aware of integration tests)
-                // see more: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators
-                return new HashSet<string>() { PropagatorTypes.Datadog };
+                return new HashSet<string>() { PropagatorTypes.B3 };
```
pellared, line 765: Update the comment for `Propagators` field.

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
```diff
@@ -24,7 +26,11 @@ public class OpenTracingTracerTests
 
         public OpenTracingTracerTests()
         {
-            var settings = new TracerSettings();
+            var settings = new TracerSettings
+            {
+                // force to be a Datadog for unit tests purposes, default is B3
+                Propagators = new System.Collections.Generic.HashSet<string> { "Datadog" }
```
pellared, line 32: How about using `PropagatorTypes.Datadog` instead of a string?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
Kielek: fixed

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
Kielek: ok, fixed

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
```diff
@@ -24,7 +27,11 @@ public class OpenTracingTracerTests
 
         public OpenTracingTracerTests()
         {
-            var settings = new TracerSettings();
+            var settings = new TracerSettings
+            {
+                // force to be a Datadog for unit tests purposes, default is B3
+                Propagators = new System.Collections.Generic.HashSet<string> { PropagatorTypes.Datadog }
```
Kielek, line 33: new jira item created to change Propagator to B3.

pellared (approved):

dszmigielski (approved):

RassK (approved):

pellared (approved):

PR #223, opened 2021-11-24
Change second profiler ID to newly generated one

dszmigielski:
## Why

There was second profiler ID introduced in upstream. We don't want to use the same IDs.

## What

Changing the introduced profiler ID to newly generated one, that is unique for us.

## Tests

We are not using it for anything for now, so I couldn't come up with any reasonable way to test it.


Kielek (approved):

pellared (approved):
I think previously it was used in more places, but I cannot find it anymore so I guess it was refactored in upstream 😉

pellared:
> We are not using it for anything for now, so I couldn't come up with any reasonable way to test it.

Try changing the `CORECLR_PROFILER` and `COR_PROFILER` values to `{0F171A24-3497-4B05-AE6D-B6B313FBF83B}` in `dev/envvars.sh` and check if the instrumentation is still working.

dszmigielski:
> > We are not using it for anything for now, so I couldn't come up with any reasonable way to test it.
> 
> Try changing the `CORECLR_PROFILER` and `COR_PROFILER` values to `{0F171A24-3497-4B05-AE6D-B6B313FBF83B}` in `dev/envvars.sh` and check if the instrumentation is still working.

Works :)

PR #10, opened 2020-03-23
Update ES span content to include db.statement

rmfitzpatrick:
These changes introduce the `SIGNALFX_INSTRUMENTATION_ELASTICSEARCH_TAG_QUERIES` environment variable to allow tagging `db.statement` from query post data.  Also updates the operation name taken from the source request parameters to be more informative.

PR #229, opened 2021-11-30
PostData as db.statement for Elasticsearch

Kielek:
Based on #10

## Why

Add missing feature from the main branch.

## What

Allow to PostData send as `db.statement` tag for Elasticsearch autoinstrumentation.
Based on #10.

There is one significant differences. Our main supports CallSite calls. Here I am adding CallTarget.
CallSite will be fully removed in scope of the next upstrean sync.

## Tests

Integration test extended based on #10.
 
Before changes: https://app.signalfx.com/#/apm/traces/01cc76d56c9534084c97ca2aec682301
After changes: https://app.signalfx.com/#/apm/traces/15b7817ab6707a35288fa796f90ea4b9


tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +86,136 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+            return true;
+        }
+
+        /// <summary>
+        /// Will replace all SanitizePatterns with Replacement, if any matches.
+        /// Used to prevent leaking sensitive information.
+        /// </summary>
+        /// <param name="data">The data to sanitize.</param>
+        /// <returns>The sanitized data.</returns>
+        public static string SanitizePostData(string data)
```
RassK, line 126: This should not be needed, as all attribute sanitization should move to the Collector. There is a single unified way how to update / delete / etc.

RassK:
> Before changes: https://app.signalfx.com/#/apm/traces/3738ffcd1f40fea1723a7b2baed0c17c
> After changes: https://app.signalfx.com/#/apm/traces/15b7817ab6707a35288fa796f90ea4b9

These seems to be similar, I can see no differences?

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
Kielek: I can remove it, could you please confirm that it will handle this things correctly?
`{"password":"realPassword"}` -> `{"password":"?"}`
or in similar manner?

Kielek:
> These seems to be similar, I can see no differences?

Wrong links in before section: https://app.signalfx.com/#/apm/traces/01cc76d56c9534084c97ca2aec682301 (first comment upadted)

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
Kielek: Fixed, I saw the description how it is handled on Collector side.

RassK: For SQL statements, it's also handled in the BE side. But I'm still missing info, how and what in detailed.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+            if (writtenBytes.Length > 1024)
+            {
+                postData = System.Text.Encoding.UTF8.GetString(writtenBytes, 0, 1024);
+            }
+            else
+            {
+                postData = System.Text.Encoding.UTF8.GetString(writtenBytes);
+            }
+
+            span.SetTag(Tags.DbStatement, postData);
```
RassK, line 140: wondering if this should / can be trimmed of excessive white-space ?

Kielek: I do not think that it is worth to do it. formatting is done on our UI. The tags are dense.

db.statements from Elasticsearch5 example
`{}
{}
{"actions":[{"add":{"index":"test_index_1","alias":"test_index_2"}}]}
{}
{}
{}
{"source":{"index":"some_index","size":10,"remote":{"host":"http://localhost:9205","username":"test_user_1","password":"supersecret"}},"dest":{"index":"some_index_clone"}}
{}
{}
{"actions":[{"add":{"index":"test_index_1","alias":"test_index_2"}}]}
{}
{}
{}
{"source":{"index":"some_index","size":10,"remote":{"host":"http://localhost:9205","username":"test_user_1","password":"supersecret"}},"dest":{"index":"some_index_clone"}}
{"create":{"_type":"post","_id":"1"}}
{"id":1,"title":"BulkCreateOperation"}
{"id":2,"title":"Create"}
{"query":{"match_all":{}}}
{}
{"create":{"_type":"post","_id":"1"}}
{"id":1,"title":"BulkCreateOperation"}
{"id":2,"title":"Create"}
{"query":{"match_all":{}}}
{}
{"shard":0}
{}
{}
{"shard":0}
{}
{}
{}
{}
{"password":"supersecret"}
{}
{}
{"password":"supersecret"}`

RassK: Currently in UI, where I was referring > 
image

Kielek: screen from the jaeger: 
image

RassK: Why our back-end can't do that 😕

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+        /// Attempts to load the PostData.WrittenBytes property from Elasticsearch.Net.RequestData.
+        /// This will return false if tagging Elasticsearch queries is disabled or if data isn't applicable
+        /// for the http method (no PostData was supplied to the request), and true otherwise.  It will
+        /// also obtain the PostData object in case manually writing the data is necessary (direct streaming enabled).
+        /// </summary>
+        /// <param name="requestData">The request data.</param>
+        /// <param name="postData">The PostData property the request data.</param>
+        /// <param name="writtenBytes">The WrittenBytes property of the PostData.</param>
+        /// <returns>Whether the request is appplicable for retrieving PostData content.</returns>
+        public static bool ShouldAttemptWrittenBytes(object requestData, out object postData, out byte[] writtenBytes)
```
pjanotti, line 98: Does it need to be public? Same Q for some other new methods below.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+        {
+            postData = null;
+            writtenBytes = null;
+            if (!Tracer.Instance.Settings.TagElasticsearchQueries)
```
pjanotti, line 102: In principle I prefer this check to be the first thing on the `SetDbStatementFromRequestData` method: kind of pay for play and avoid two calls for that scenario in which it was turned off.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+        /// <param name="writtenBytes">The byte[] of data with which to encode, sanitize, truncate, and tag.</param>
+        public static void SetDbStatement(Span span, byte[] writtenBytes)
+        {
+            if (span == null || writtenBytes == null)
+            {
+                return;
+            }
+
+            string postData = null;
+            if (writtenBytes.Length > 1024)
```
pjanotti, line 131: I understand that here it is doing the trimming before building the string so it saves some work before reaching the exporter that has a trimming too. However, we should use the same configuration setting to control both.

Add a comment to justify this "early" trimming.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+        /// <param name="methodName">The method to retrieve from PostData.</param>
+        /// <param name="postData">The PostData from which to retrieve the desired MethodInfo.</param>
+        /// <param name="requestData">The RequestData from which to retrieve the ConnectionSettings used by the write method.</param>
+        /// <param name="connectionSettings">The RequestData.ConnectionSettings property.</param>
+        /// <returns>The Write or WriteAsync MethodInfo.</returns>
+        public static MethodInfo GetWriteMethodInfo(string methodName, object postData, object requestData, out object connectionSettings)
+        {
+            connectionSettings = requestData.GetProperty("ConnectionSettings").GetValueOrDefault();
+            var postDataType = postData.GetType();
+            return postDataType.GetMethod(methodName);
```
pjanotti, line 157: Just wondering if it is worth to cache these methods, not sure if it is worth the trouble. Perhaps, just add it to the backlog.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+        /// ConnectionSettings used by these methods.
+        /// </summary>
+        /// <param name="methodName">The method to retrieve from PostData.</param>
+        /// <param name="postData">The PostData from which to retrieve the desired MethodInfo.</param>
+        /// <param name="requestData">The RequestData from which to retrieve the ConnectionSettings used by the write method.</param>
+        /// <param name="connectionSettings">The RequestData.ConnectionSettings property.</param>
+        /// <returns>The Write or WriteAsync MethodInfo.</returns>
+        public static MethodInfo GetWriteMethodInfo(string methodName, object postData, object requestData, out object connectionSettings)
+        {
+            connectionSettings = requestData.GetProperty("ConnectionSettings").GetValueOrDefault();
```
pjanotti, line 155: In principle, GetProperty can return null. If we don't need to check for null in this case please add a comment.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
```diff
@@ -79,5 +84,113 @@ public static Scope CreateScope(Tracer tracer, IntegrationInfo integrationId, st
...
+            byte[] writtenBytes;
+            if (!ShouldAttemptWrittenBytes(requestData, out postData, out writtenBytes))
+            {
+                return;
+            }
+
+            if (writtenBytes == null)
+            {
+                object connectionSettings;
+                var methodInfo = GetWriteMethodInfo("Write", postData, requestData, out connectionSettings);
```
pjanotti, line 184: Are you planning to re-use this method in other place? I only saw it being used here to get the "Write" method.

Kielek: Leftover from migration from main. Method signature simplified.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
Kielek: I think that this value was set arbitrary. Just to take the first part of the information.
Max length for the tag is taken from `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` and by defult is equal to 12000.

I created separate item to change it if needed.

1024 const extracted and commented that it is arbitrary value.

tracer/src/Datadog.Trace/ClrProfiler/Integrations/Elasticsearch/ElasticsearchNetCommon.cs
RassK: When refactoring reflection based logic, it is better to build expressions and cache these.

RassK (approved):

dszmigielski (approved):

pjanotti (approved):
LGTM

PR #231, opened 2021-11-30
Update span name in tracing module

lachmatt:
## Why
In order to have good, low-cardinality names for spans created in `TracingHttpModule`.

## What
Update name for spans created in `TracingHttpModule` with resource names created based on routes.

On _main_ there is flag (i.e `UseWebServerResourceAsOperationName`) that controls whether span name should be overriden.
However, we already changed names for spans created in several integrations without checking flag's value. 
On _main_, resourceName created based on routing information is passed inside `HttpContext` to `TracingHttpModule` only if `RouteTemplateResourceNamesEnabled` flag is set(which is by default). There is useful information (i.e resourceName) to be passed even if it's disabled, that's why I dropped that check in `AspNetMvcIntegration` and `AspNetWebApi2Integration`.

## Tests
Tests with verification files (aspnetmvc4 for now)


RassK (commented):
Just breaking changes changelog is missing, that the `UseWebServerResourceAsOperationName` is removed. (and what's the new fixed behavior)

pjanotti (approved):
LGTM. Per our conversation it would be nice to be have a consistent way to map requests using default "controller.action" to a single path that expresses the same even if the actual `url` is different. That said this is already an improvement and anything else can come in a separate PR.

RassK (approved):

PR #235, opened 2021-12-01
Fix RabbitMQ and Kafka span names

RassK:
## What

Unifies messaging systems (RabbitMQ, Kafka) span naming according to the OTel rules. Uses `(default) <OperationName>` if topic / queue name is missing. Similar as OTel `(temporary)` if topic / queue is not a constant. 

## Tests

Before changes: https://app.signalfx.com/#/apm/traces/3fcf26d05948ddec537d74474fc1377a
After changes: https://app.signalfx.com/#/apm/traces/40505b5bde05e4f90e41694fca8eee42


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -34,16 +36,18 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
 
                 var parent = tracer.ActiveScope?.Span;
                 if (parent is not null &&
-                    parent.OperationName.StartsWith(KafkaConstants.ProduceOperationName) &&
-                    parent.GetTag(Tags.InstrumentationName) != null)
+                    parent.GetTag(Tags.InstrumentationName) != null &&
+                    ReferenceEquals(parent.LogicScope, KafkaConstants.ProduceOperationName))
```
RassK, line 40: @dszmigielski I think I was wrong here when referenced to propagation. Seems it's protecting that there aren't 2 spans for exactly the same context (in-proc).

lachmatt (approved):

pjanotti (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -101,7 +106,7 @@ internal static Scope CreateProducerScope(Tracer tracer, ITopicPartition topicPa
 
                 var parent = tracer.ActiveScope?.Span;
                 if (parent is not null &&
-                    parent.OperationName.StartsWith(KafkaConstants.ConsumeOperationName) &&
+                    ReferenceEquals(parent.LogicScope, KafkaConstants.ConsumeOperationName) &&
```
pellared, line 109: What is the reason we are using `ReferenceEquals` and not `==`? Optimization?

pellared (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
RassK: for the future > [sharplab.io](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgMwAEOZAwmQN5FmMXnAD2rANmQLIAUmuAAxkIcCkLLAAlHSZyyDeY0wB2EWQC8GyQG5FjAL76FhOaUnsuAOX4TR44dNlKTLimoBKAUwBmXgE5eMADGXgCiAI4ArhAcKLz20nqmTEaEBkA=)

PR #240, opened 2021-12-07
Fix NLog integration when using ILogger

pjanotti:
## Why

When NLog is used as the provider for ILogger the current optimization is broken: the values are reset after the first execution context is gone and become empty.

## What

Instead of leveraging the async local from the scope the code now subscribe to the typical events used for log correlation. However, instead of using the "context dictionary" used in the typical implementation it simply adds the values to the mapped logical context. It is likely that the same could be done for the log providers but I didn't have the time to validate if the same optimization can be applied to other log providers.

## Tests

before changes:

* Validated that the log correlation was broken in a ASP.NET Core app that used ILogger.

after changes:

* Validated that the log correlation was fixed in a ASP.NET Core app that used ILogger.
* Promoted the NLog version used in tests from 4.5.11 to 4.6.8: the test can't pass for both versions without an explicit version check so opting for upgrading the test for the latest 4.6.* version. However, I manually validated that log correlation also works for 4.5.11

## Performance Impact
As expected the code is not as fast as the previous optimization but it is better than the code using the "context dictionary".  *OldTagging* in the table below is the code using the "context dictionary".

|                     Method |         Categories |      Mean |     Error |    StdDev |    Median | Ratio | RatioSD |  Gen 0 |  Gen 1 |  Gen 2 | Allocated |
|--------------------------- |------------------- |----------:|----------:|----------:|----------:|------:|--------:|-------:|-------:|-------:|----------:|
|                  NoTagging | NoActiveSpanSwitch |  2.734 us | 0.0436 us | 0.0340 us |  2.724 us |  1.00 |    0.00 | 0.5798 | 0.2899 | 0.0381 |    2.4 KB |
|                 OldTagging | NoActiveSpanSwitch |  9.532 us | 0.1831 us | 0.1712 us |  9.568 us |  3.48 |    0.08 | 1.4648 | 0.3510 | 0.0763 |   8.34 KB |
|           NewNLog46Tagging | NoActiveSpanSwitch |  7.186 us | 0.1437 us | 0.1345 us |  7.142 us |  2.63 |    0.07 | 1.0834 | 0.3510 | 0.0763 |   6.45 KB |
|                            |                    |           |           |           |           |       |         |        |        |        |           |
|        NoTaggingSpanSwitch |   ActiveSpanSwitch |  4.785 us | 0.0949 us | 0.2123 us |  4.816 us |  1.00 |    0.00 | 0.5951 | 0.3052 | 0.0458 |   3.53 KB |
|       OldTaggingSpanSwitch |   ActiveSpanSwitch | 18.199 us | 0.6247 us | 1.8222 us | 17.614 us |  3.73 |    0.41 | 3.7231 | 0.8240 | 0.3357 |  15.07 KB |
| NewNLog46TaggingSpanSwitch |   ActiveSpanSwitch | 13.734 us | 0.2622 us | 0.5051 us | 13.769 us |  2.88 |    0.20 | 2.1210 | 0.4883 | 0.1526 |  11.38 KB |

### DIFF
--- a/src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
+++ b/src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
@@ -55,7 +55,7 @@ namespace SignalFx.Tracing.Logging
                     _scopeManager.SpanOpened += StackOnSpanOpened;
                     _scopeManager.SpanClosed += StackOnSpanClosed;
             }
-            else if (_logProvider is NLogLogProvider)
+            else if (_logProvider is NLogLogProvider && UseNLogOptimized())
             {
                 // This NLog version can use the value readers optimization.
                 _scopeManager.SpanActivated += ActivateMappedContext;
@@ -97,7 +97,7 @@ namespace SignalFx.Tracing.Logging
                 _scopeManager.SpanOpened -= StackOnSpanOpened;
                 _scopeManager.SpanClosed -= StackOnSpanClosed;
             }
-            else if (_logProvider is NLogLogProvider)
+            else if (_logProvider is NLogLogProvider && UseNLogOptimized())
             {
                 _scopeManager.SpanActivated -= ActivateMappedContext;
                 _scopeManager.TraceEnded -= DeactivateMappedContext;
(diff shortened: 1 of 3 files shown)


src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
```diff
@@ -53,10 +55,11 @@ public LibLogScopeEventSubscriber(IScopeManager scopeManager, TracerSettings set
                     _scopeManager.SpanOpened += StackOnSpanOpened;
                     _scopeManager.SpanClosed += StackOnSpanClosed;
             }
-            else if (_logProvider is NLogLogProvider && UseNLogOptimized())
+            else if (_logProvider is NLogLogProvider)
             {
                 // This NLog version can use the value readers optimization.
-                UseValueReaders(settings);
+                _scopeManager.SpanActivated += ActivateMappedContext;
```
lachmatt, line 61: I think this breaks for NLog1.0 (CI seems to confirm that). `LogProvider.OpenMappedContext` call should be wrapped with `try/catch` block

pjanotti: Yeah, I didn't not run that test locally. While we don't support NLog 1.0 for log correlation we don't want to break it.

pjanotti:
Test failures seem unrelated, kicking it again. The following test seems flaky on Windows: `Datadog.Trace.ClrProfiler.IntegrationTests.AdoNet.SqlCommandTests.SubmitsTraces(packageVersion: "")`

PR #249, opened 2021-12-13
Vendor protobuf

Kielek:
## Why

To be ready to send metrics via protobuff protocol.

## What

Serialization classed for sending metrics via protobuff protocol.

## Tests

App created locally with possibility to serialize and deserialize DataPointUploadMessage to/from file.


tracer/dependabot/Datadog.Dependabot.Vendors.csproj
```diff
@@ -21,6 +21,9 @@
     <!-- https://www.nuget.org/packages/Newtonsoft.Json/12.0.1 -->
     <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
 
+    <!-- https://www.nuget.org/packages/protobuf-net/2.4.0 -->
```
pellared, line 24: first of all we need be aware that it does not work right now

> To configure Dependabot, you must use this repository’s default branch.

even if it would be enabled it would not work for us because of

https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/.github/dependabot.yml#L13

What do you think of creating a separate directory+csproj for "our" vendored dependencies and updating `/.github/dependabot.yml` accordingly?

Why `2.4.0` and not the latest or at least `2.4.6`? If it is as desired then I would leave here a comment with an explanation

Kielek: This file is extended by default for all Vendored libraries. I do not think that it is worth the extract it to separate file.
It could introduce additional complication while upstream sync for build tasks. What is more, I do not expect that dependabot will handle it.

protobuf-net is updated to 2.4.6.

pellared (approved):

PR #254, opened 2021-12-14
Initial metric exporter implementation

lachmatt:
## Why

Export metrics.

## What
Simple implementation of metrics exporter:

- No buffering/batching before sending
- Sync send
- `SignalFxReporter` and its dependencies e.g `WebRequestor`, `WebRequestorFactory` migrated from dotnet lambda wrapper repository (I plan to simplify that part of the code)

## Tests
Automated test for `SignalFxMetricSender`.
Smoke tests with backend. https://app.signalfx.com/#/temp/chart/FGkUY9SAcAE


Kielek:
Do you see any reasonable way to avoid manual code modification under Vendored folder?
It will be overwritten if you execute nuke `UpdateVendoredCode` task.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -179,6 +179,12 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.AgentUri"/>
         public const string AgentUri = "SIGNALFX_TRACE_AGENT_URL";
 
+        /// <summary>
+        /// Configuration key for the URL where the Tracer can send metrics.
+        /// </summary>
+        /// <seealso cref="TracerSettings.MetricsAgentUri"/>
+        public const string MetricsAgentUri = "SIGNALFX_METRICS_AGENT_URL";
```
Kielek, line 186: It will be great to put both new keys to readme. Even with simple description.

pellared:
> Smoke tests with backend.

I would love to have a hyperlink and info what was tested - even for sake of knowledge-sharing 😉

tracer/src/Datadog.Trace/SignalFx/Metrics/MetricsExporterType.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Enumeration for the available metric exporter types.
+    /// </summary>
+    public enum MetricsExporterType
+    {
+        /// <summary>
+        /// StatsD based exporter.
+        /// </summary>
+        Default,
```
Kielek, line 11: For our purposes, SignalFx should be default.
Consider one of two defaulting options, like `ExporterType` or `PropagatorTypes`.

lachmatt:
> Do you see any reasonable way to avoid manual code modification under Vendored folder? It will be overwritten if you execute nuke `UpdateVendoredCode` task.

I looked into it, I think it might be reduced (sometimes at the cost of readability imo) but not sure if avoided.

lachmatt:
> > Smoke tests with backend.
> 
> I would love to have a hyperlink and info what/how it was tested - even for sake of knowledge-sharing 😉

Added

pellared:
> Do you see any reasonable way to avoid manual code modification under Vendored folder? It will be overwritten if you execute nuke `UpdateVendoredCode` task.

I looked at it and I do not see any easy way without introducing a lot of complexity. I suggest waiting to see if it would be an issue. Then we may tackle it with more insight what are the potential conflicts and other problems.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
pellared: How about renaming to `SIGNALFX_METRICS_ENDPOINT_URL` to have it similar to `SIGNALFX_ENDPOINT_URL`?

tracer/src/Datadog.Trace/Vendors/StatsdClient/IDogStatsd.cs
```diff
@@ -22,7 +23,8 @@ internal interface IDogStatsd : IDisposable
         /// Must be called before any other methods.
         /// </summary>
         /// <param name="config">The value of the config.</param>
-        void Configure(StatsdConfig config);
+        /// <param name="settings">The value of settings.</param>>
+        void Configure(StatsdConfig config, ImmutableTracerSettings settings);
```
pellared, line 27: I think we do not need to change the interface. We could inject this data via constructor instead. `Configure` is used only in `TraceManagerFactory`. It should simplify the code and reduce the amount of changes.

lachmatt:
Updated not to modify vendored code

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricSender.cs
```diff
@@ -0,0 +1,61 @@
...
+using MetricType = Datadog.Trace.Vendors.StatsdClient.MetricType;
+
+namespace Datadog.Trace.SignalFx.Metrics
+{
+    internal class SignalFxMetricSender
+    {
+        private readonly ISignalFxReporter _reporter;
+        private readonly string[] _globalTags;
+
+        public SignalFxMetricSender(ISignalFxReporter reporter, string[] globalTags)
```
pellared, line 13: I would probably name the classes the other way as the `Sender` does not send anything 😄 

How about changing:

- `SignalFxMetricSender` -> `SignalFxMetricsReporter` or `SignalFxMetricsExporter`
- `SignalFxReporter` -> `SignalFxMetricSender`

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -179,6 +179,12 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.AgentUri"/>
         public const string AgentUri = "SIGNALFX_TRACE_AGENT_URL";
 
+        /// <summary>
+        /// Configuration key for the URL where the Tracer can send metrics.
+        /// </summary>
+        /// <seealso cref="TracerSettings.MetricsAgentUri"/>
+        public const string MetricsAgentUri = "SIGNALFX_METRICS_ENDPOINT_URL";
```
pellared, line 186: nit: can we rename this field? e.g. to `MetricsEndpoint` or `MetricsEndpointUrl` ?

tracer/src/Datadog.Trace/Configuration/ImmutableTracerSettings.cs
```diff
@@ -120,6 +123,12 @@ public ImmutableTracerSettings(TracerSettings settings)
         /// <seealso cref="ConfigurationKeys.AgentPort"/>
         public Uri AgentUri { get; }
 
+        /// <summary>
+        /// Gets the Uri where the Tracer can connect to the metrics Agent.
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.MetricsAgentUri"/>
+        public Uri MetricsAgentUri { get; }
```
pellared, line 130: nit: can we rename this field? e.g. to `MetricsEndpoint` or `MetricsEndpointUrl` ?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -156,6 +156,14 @@ public TracerSettings(IConfigurationSource source)
                             // default value
                             8125;
 
+            // TODO: revisit the default value
```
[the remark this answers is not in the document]
pellared: would be good to update `otel-config.yaml` (can be done in a separate PR) - maybe worth creating a Jira story to not forget about it.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricSender.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            _reporter = reporter ?? throw new ArgumentNullException(nameof(reporter));
+            _globalTags = globalTags ?? throw new ArgumentNullException(nameof(globalTags));
+        }
+
+        public void SendMetric(MetricType metricType, string name, double value, string[] tags = null)
+        {
+            if (metricType == MetricType.Gauge || metricType == MetricType.Count)
```
pellared, line 21: Worth to add a comment that only sends metrics for `Gauge` and `Count` types.
However, I propose to split this into two methods `SendGaugeMetric` and `SendCountMetric` it will be more obvious then and it will require less conditions.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxStats.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using Datadog.Trace.DogStatsd;
+using Datadog.Trace.Vendors.StatsdClient;
+
+namespace Datadog.Trace.SignalFx.Metrics
+{
+    /// <summary>
+    /// Supports sending counter and gauge metrics with provided sender.
+    /// Other metrics/events are ignored.
```
pellared, line 9: Can you explain why we are ignoring them?

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxStats.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    internal class SignalFxStats : IDogStatsd
+    {
+        private readonly SignalFxMetricSender _metricSender;
+
+        public SignalFxStats(SignalFxMetricSender metricSender)
+        {
+            _metricSender = metricSender ?? throw new ArgumentNullException(nameof(metricSender));
```
pellared, line 17: How about a little refactor to depend on:
- `SignalFxMetricConverter` -> this could be even a static class :smile:
- `ISignalFxMetricSender` -> this is what is currently `ISignalFxReporter` is doing

This class would then first convert the payload and then send it on the wire.

Then the tests for `SignalFxMetricConverter` would look nicer than for current version of `SignalFxMetricSender`

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
```diff
@@ -0,0 +1,90 @@
...
+        private readonly WebRequestorFactory _requestorFactory;
+
+        public SignalFxReporter(Uri metricsEndpointAddress, string apiToken)
+        {
+            if (metricsEndpointAddress == null)
+            {
+                throw new ArgumentNullException(nameof(metricsEndpointAddress));
+            }
+
+            _requestorFactory = new WebRequestorFactory()
```
pellared, line 28: 1. Why are we not using `HttpClient`?
2. Not sure if we need a builder here. It makes reading better, but it costs a little performance and personally I think I never head issues with `HttpClient` API.

docs/advanced-config.md
```diff
@@ -63,6 +63,7 @@ Use these environment variables to configure the tracing library:
 | `SIGNALFX_TRACE_ENABLED` | Enable to activate the tracer. | `true` |
 | `SIGNALFX_TRACE_DEBUG` | Enable to activate debugging mode for the tracer. | `false` |
 | `SIGNALFX_ENDPOINT_URL` | The URL to where trace exporters send traces. | `http://localhost:8126` |
+| `SIGNALFX_METRICS_ENDPOINT_URL` | The URL to where metric exporters send metrics. | `http://localhost:9943` |
```
Kielek, line 66:
```suggestion
| `SIGNALFX_METRICS_ENDPOINT_URL` | The URL to where metric exporters send metrics. | `http://localhost:9943/v2/datapoint` |
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -156,6 +156,13 @@ public TracerSettings(IConfigurationSource source)
                             // default value
                             8125;
 
+            var metricsUri = source?.GetString(ConfigurationKeys.MetricsAgentUri) ??
+                             // default value
+                             "http://localhost:9943";
```
Kielek, line 161:
```suggestion
                             "http://localhost:9943/v2/datapoint";
```
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/078d6bd81584f8010d10e929584445a4f4e55f01/receiver/signalfxreceiver/receiver.go#L142
Probably also modification in our dev/docker-compose is needed.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
lachmatt: This code was migrated from lambda wrapper repository as suggested.
I agree builder is not needed, I plan to simplify this code (mentioned that in PR's description).

tracer/src/Datadog.Trace/SignalFx/Metrics/WebRequestor.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public Stream GetWriteStream()
+        {
+            return _request.GetRequestStream();
+        }
+
+        public void Send()
+        {
+            using var resp = (HttpWebResponse)_request.GetResponse();
```
pellared, line 42: AFAIK it usally throws exception for non 2xx HTTP status codes. See:
- https://stackoverflow.com/questions/1366848/httpwebrequest-getresponse-throws-webexception-on-http-304
- https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.getresponse?view=net-6.0
- https://docs.microsoft.com/en-us/dotnet/api/system.net.webexception?view=net-6.0

What will happen if this code throws `WebException` ? PS. I do not like the `WebRequest` API :wink:

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+
+                try
+                {
+                    requestor.Send();
+                    Log.Debug($"Sent {msg.datapoints.Count} metrics");
+                }
+
+                // TODO: handle exceptions similarly to ZipkinExporter
+                catch (SecurityException)
```
pellared, line 67: I think that this code would never be reached. AFAIK `WebException` would be thrown by `_requestorFactory.GetRequestor` for `403` HTTP Status code.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public void Send(DataPointUploadMessage msg)
+        {
+            if (msg == null)
+            {
+                throw new ArgumentNullException(nameof(msg));
+            }
+
+            try
```
pellared, line 50: `try` in `try` is ugly here. We should refactor to make the `try`'s on the same level.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricSender.cs
pellared: I think this proposal is better: https://github.com/signalfx/signalfx-dotnet-tracing/pull/254#discussion_r770315674

docs/advanced-config.md
pellared: This config file is "official" documentation. We should:
1. Mention it in the `CHANGELOG.md` if we want to start supporting it in the upcoming release. Also I think we would need to add `SIGNALFX_RUNTIME_METRICS_ENABLED` here and mention it as well.
2. Create a section in `internal-config.md` like `Unpublished settings`
2.

dev/otel-config.yaml
```diff
@@ -48,9 +52,9 @@ service:
       processors: [batch]
       exporters: [logging, jaeger, sapm]
     metrics:
-      receivers: [otlp, opencensus]
+      receivers: [signalfx]
       processors: [batch]
-      exporters: [prometheus]
+      exporters: [signalfx]
```
Kielek, line 57: logging, separate PR?

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
lachmatt: fixed

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxReporter.cs
lachmatt: fixed

docs/advanced-config.md
lachmatt: fixed, created new section as suggested

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxStats.cs
lachmatt: discussed offline, agreed to leave it as it is

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricSender.cs
lachmatt: done

Kielek (approved):

PR #255, opened 2021-12-14
Update title and summary

pellared:
## Why

The title and summary feel legacy.

## What

1. Update the name to reflect that it is not only about tracing (metrics and profiling via logs is coming)
2. Update the summary to describe better the current state

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -31,7 +31,7 @@ and [`ActivitySource`](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/
 
 ## Requirements
 
-## Supported .NET versions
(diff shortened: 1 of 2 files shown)


theletterf (approved):
@ivomagi Still hesitant to use "SignalFx" instead of "Splunk" but I'll defer to your decision.

Kielek (approved):

dszmigielski (approved):

ivomagi:
From my side I tend to prefer to keep the SignalFX label on the name, staying in sync with all the other components that are not yet on OpenTelemetry. Expanding the scope to be clearer on "this not just tracing" is correct.

PR #251, opened 2021-12-13
Documentation restructure

pellared:
## Why

Restructure the docs to make it similar to https://github.com/signalfx/splunk-otel-java and https://github.com/signalfx/splunk-otel-python. Use https://github.com/signalfx/gdi-specification/tree/main/specification/templates

## What

See commit messages.
 
## Next PRs

- create "sections" for settings in advanced configuration
- create a dedicated section for instrumenting an application deployed to IIS
- simplify the `Get started` for Windows. some stuff should go to advanced configuration
- automating `Get started` for Windows also PowerShell
- document that Splunk OTel Collector instance must be up
- update `DEVELOPING.md` (nuke instructions, VS components etc)
- address other feedback and findings

PR #261, opened 2021-12-15
Describe SIGNALFX_TRACE_SAMPLE_RATE default value

pellared:
## Why

Address https://github.com/signalfx/signalfx-dotnet-tracing/pull/251#discussion_r767689600

## What

Describe what happens when no value is set. From user perspective it looks like it would be almost the same as if the user would provide `1.0`. However if user provides a value then some structures (samplers) are created so there could be some little overhead...  This is why I have not written that the default is `1.0`.

## Tests

None. Reading code.


docs/advanced-config.md
```diff
@@ -52,7 +52,7 @@ Use these environment variables to configure the tracing library:
 | `SIGNALFX_TRACE_RESPONSE_HEADER_ENABLED` | Enable to add server trace information to HTTP response headers. | `true` |
 | `SIGNALFX_TRACE_STARTUP_LOGS` | Enable to activate diagnostic log at stratup. | `true` |
 | `SIGNALFX_TRACE_SAMPLING_RULES` | Comma separated list of sampling rules taht enabled custom sampling rules based on regular expressions. The rule is matched in order of specification. The first match in a list is used. The item "sample_rate" is required in decimal format. The item "service" is optional in regular expression format, to match on service name. The item "name" is optional in regular expression format, to match on operation name. | `'[{"sample_rate":0.5, "service":"cart.*"}],[{"sample_rate":0.2, "name":"http.request"}]'` |
-| `SIGNALFX_TRACE_SAMPLE_RATE` | The global rate for the sampler. |  |
+| `SIGNALFX_TRACE_SAMPLE_RATE` | The global rate for the sampler. By default, all traces are sampled.  |  |
```
Kielek, line 55:
```suggestion
| `SIGNALFX_TRACE_SAMPLE_RATE` | The global rate for the sampler. By default, all traces are sampled. |  |
```

Kielek (approved):

PR #270, opened 2021-12-16
Initial work on thread sampling

johnbley:
Initial work on thread sampling:
- off by default (turn on with `SIGNALFX_THREAD_SAMPLING_ENABLED=true`)
- 1 sec sampling period by default, configurable
- stub in managed code that receives thread samples currently only prints to stdout
- No checked in benchmarks yet; no unit tests in C# (some in C++)


pellared:
I see that there is some problem that the `ci` GitHub workflow is not executed

Probably [this](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request) is the reason: 
> Workflows will not run on pull_request activity if the pull request has a merge conflict. The merge conflict must be resolved first.

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
```diff
@@ -121,6 +121,10 @@ namespace environment {
     // Sets whether to enable NGEN images.
     const WSTRING clr_enable_ngen = WStr("SIGNALFX_CLR_ENABLE_NGEN");
 
+    // If you change this, change corresponding logic in Instrument.cs too
+    const WSTRING thread_sampling_enabled = WStr("SIGNALFX_THREAD_SAMPLING_ENABLED");
```
pellared, line 125: how about changing it to `SIGNALFX_PROFILER_ENABLED`?
reason: https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md

it would be good also to document the env var in `docs/internal-config.md` in some new section named e.g. `Under development` or `Not published` or whatever naming would be better 😉

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
```diff
@@ -121,6 +121,10 @@ namespace environment {
...
+
+    const WSTRING thread_sampling_period = WStr("SIGNALFX_THREAD_SAMPLING_PERIOD");
```
pellared, line 127: how about changing it to `SIGNALFX_PROFILER_CALL_STACK_INTERVAL`?
reason: https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md

it would be good also to document the env var in `docs/internal-config.md` in some new section named e.g. `Under development` or `Not published` or whatever naming would be better 😉

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
johnbley: We (and the CLR) already have a bunch of `_PROFILER_` env variables so I wanted to be super clear about which feature this was controlling.  We can iron out all these names as a separate PR.

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
johnbley: See above - let's iron out naming separately from this initial work.
Docs - yes, should do that in the next PR.  Just trying to get the legs under this thing right now.

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
pellared: Sure

Kielek (approved):

Kielek:
I have created bunch of stories in our jira to cover some issues in the code.
As the code will be not executed by default, I would merge it, then handle stories one by one. It will be easier to discuss changes in smaller chunks.

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -0,0 +1,659 @@
...
+#include <map>
+
+#define MAX_FUNC_NAME_LEN 256
+#define MAX_CLASS_NAME_LEN 512
+#define MAX_STRING_LENGTH 512
+
+#define MAX_CODES_PER_BUFFER (10 * 1000)
+
+// If you change this, consider ThreadSampler.cs too
+#define SAMPLES_BUFFER_MAXIMUM_SIZE (200 * 1024)
```
RassK, line 13: Maybe PInvoke these values? so there is only a single place to change.
(another way from managed to native is also option, see InstrumentationDefinitions.cs)

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleNativeFormatParser.cs
```diff
@@ -0,0 +1,159 @@
...
+namespace Datadog.Trace.ThreadSampling
+{
+    /// <summary>
+    /// Parser the native code's pause-time-optimized format.
+    /// </summary>
+    internal class ThreadSampleNativeFormatParser
+    {
+        private static bool printStackTraces = false;
+
+        private byte[] buf;
```
RassK, line 14: wondering that it passes code style inspection 😕 privates are with `_` prefix.

PR #287, opened 2021-12-29
refactor ThreadSampling env vars

Kielek:
## Why

Cleanup scaffolding for Thread Sampling.
Values needed for exporter.

## What

Refactor reading env. variables for Thread Sampling

## Tests

Manual test for Console application. Still working.

### DIFF
--- a/tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
+++ b/tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
@@ -3,7 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-using System;
+// Modified by Splunk Inc.
+
 using System.Threading;
 using Datadog.Trace.ClrProfiler;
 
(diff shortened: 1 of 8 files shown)


tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -180,9 +181,7 @@ public static void Initialize()
             // Thread Sampling ("profiling") feature
             try
             {
-                // If you change this, change environment_variables.h too
-                var enabled = EnvironmentHelpers.GetEnvironmentVariable("SIGNALFX_THREAD_SAMPLING_ENABLED", "false");
-                if (enabled != null && (enabled.ToLower() == "1" || enabled.ToLower() == "true"))
+                if (TracerManager.Instance.Settings.ThreadSamplingEnabled)
```
pellared, line 184: you should be able to pass this value through the constructor in `TracerManagerFactory.CreateTracerManager()`
`

pellared (commented):
just comments to reduce the "globals" explosion 😄 
apart from it LGTM

tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
```diff
@@ -12,20 +13,12 @@ namespace Datadog.Trace
...
-            var enabled = Environment.GetEnvironmentVariable("SIGNALFX_THREAD_SAMPLING_ENABLED");
-            if (enabled != null && (enabled.ToLower() == "1" || enabled.ToLower() == "true"))
-            {
-                PushScopeToNative = true;
-            }
-            else
-            {
-                PushScopeToNative = false;
-            }
+            PushScopeToNative = TracerManager.Instance.Settings.ThreadSamplingEnabled;
```
pellared, line 21: you should be able to pass this value through the constructor in `TracerManagerFactory.CreateTracerManager()`

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampler.cs
```diff
@@ -53,7 +27,7 @@ private static void ReadOneSample(byte[] buf)
 
         private static void SampleReadingThread()
         {
-            int period = GetThreadSamplingPeriod();
+            int period = TracerManager.Instance.Settings.ThreadSamplingPeriod;
```
pellared, line 30: you should be able to pass this value from `Instumentation.Initialize()` via  `ThreadSampler.Initialize()`

tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
Kielek: fixed

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampler.cs
Kielek: fixed

pellared (approved):

PR #291, opened 2021-12-30
Minor thread sampling tweaks

johnbley:
- document buffer format a bit
- clean up some warnings
- reduce size of length prefix for strings since they are limited to 1k anyways
- fix type issue when NameCache was reused for class names

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
@@ -75,7 +75,7 @@ int ThreadSampling_ConsumeOneThreadSample(int len, unsigned char* buf)
     {
         return 0;
     }
-    int toUseLen = min(toUse->size(), len);
+    int toUseLen = (int) min(toUse->size(), len);
     memcpy(buf, toUse->data(), toUseLen);
     delete toUse;
     return toUseLen;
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -152,8 +177,8 @@ void ThreadSamplesBuffer::StartBatch()
 void ThreadSamplesBuffer::StartSample(ThreadID id, ThreadState* state, ThreadSpanContext context)
 {
     CHECK_SAMPLES_BUFFER_LENGTH();
-    writeByte(0x02);
-    writeInt(id);
+    writeByte(THREAD_SAMPLES_START_SAMPLE);
+    writeInt((int)id); // FIXME not really sure how to map this to anything; needs more research
```
Kielek, line 181: Could you please create separate jira item to fix it?

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -618,7 +648,7 @@ WSTRING* NameCache::get(FunctionID key)
     return found->second->second;
 }
 
-void NameCache::put(FunctionID key, WSTRING* val)
+void NameCache::put(UINT_PTR key, WSTRING* val)
 {
     auto pair = std::pair<FunctionID, WSTRING*>(key, val);
     list.push_front(pair);
```
Kielek, line 654: any chance to fix warning also in line 657 - if (map.size() > maxSize)

Kielek (approved):

pellared (approved):
wouldn't changing the type of `NameCache.maxSize` to `size_t` fix some warnings?

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
pellared: wouldn't changing the type of `NameCache.maxSize` to `size_t` fix some warnings?

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
johnbley: (done)

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
johnbley: Good idea, done.

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -75,7 +75,7 @@ int ThreadSampling_ConsumeOneThreadSample(int len, unsigned char* buf)
     {
         return 0;
     }
-    int toUseLen = min(toUse->size(), len);
+    int toUseLen = (int) min(toUse->size(), len);
```
pellared, line 78: @johnbley I guess that `len` argument of `ThreadSampling_ConsumeOneThreadSample` should also be of type `size_t`

PR #299, opened 2022-01-05
Add Azure Web Service docs

RassK:
## What

Adds Azure Web Service instrumentation instructions. 

## Tests

Repeated steps in Azure. 
https://app.signalfx.com/#/apm/traces/533732f193fcb568c6380a6441b381f2

## Next 

Any ideas does it cover webjobs or they require something more special?


docs/azure.md
```diff
@@ -0,0 +1,10 @@
+# Azure services instrumentation guide
+
+## App Service
+
+1. Choose your App Service.
+2. Navigate to `Development Tools` > `Extensions`.
+3. Find and install `SignalFx .NET Tracing` extension.
+4. Navigate to `Settings` > `Configuration`.
+5. Add `New application setting` to configure receiver. See [advanced-config.md](advanced-config.md#exporter-settings) exporter settings section for more details.
```
Kielek, line 9: Could you please be more specific which environmental variables are mandatory?
In previous document three of them was marked in this way `SIGNALFX_SERVICE_NAME`, `SIGNALFX_ACCESS_TOKEN` and  `SIGNALFX_ENDPOINT_URL`.

pellared: And probably it is worth saying that `SIGNALFX_ACCESS_TOKEN` and `SIGNALFX_ENDPOINT_URL` should be set in a way to send the telemetry directly to the Splunk Observability Cloud.

pellared: Also, I think a step like `Restart your App Service.` is missing at the end.

docs/azure.md
```diff
@@ -0,0 +1,10 @@
+# Azure services instrumentation guide
```
pellared, line 1:
```suggestion
# Azure instrumentation guide
```

WDYT? to keep in sync with the header from `README.md`

docs/README.md
```diff
@@ -116,14 +116,18 @@ export SIGNALFX_SERVICE_NAME="my-service-name"  # Set the service name
...
+## Instrument a Windows Service running a .NET application
 
 See [windows-service-instrumentation.md](windows-service-instrumentation.md).
 
-### Instrument an ASP.NET application deployed on IIS
+## Instrument an ASP.NET application deployed on IIS
 
 See [iis-instrumentation.md](iis-instrumentation.md).
 
+## Azure instrumentation guide
```
pellared, line 127:
```suggestion
## Instrument Azure App Service
```

the header would be similar to other headers

also, it is similar to: https://docs.signalfx.com/en/latest/apm/apm-instrument/apm-azure-app.html

docs/azure.md
```diff
@@ -0,0 +1,10 @@
...
+
+## App Service
```
pellared, line 3:
```suggestion
# Instrument Azure App Service
```

to have the same title as the header in `README.md`

docs/azure.md
```diff
@@ -0,0 +1,10 @@
...
+
+1. Choose your App Service.
+2. Navigate to `Development Tools` > `Extensions`.
+3. Find and install `SignalFx .NET Tracing` extension.
```
pellared, line 7:
```suggestion
3. Find and install the `SignalFx .NET Tracing` extension.
```

docs/azure.md
Kielek: As I remember, `SIGNALFX_ACCESS_TOKEN` is mandatory. Otherwise, profiling for Azure App Service won't work.
We can change it - requierments from upstream.

pjanotti: @RassK Include setting `SIGNALFX_AZURE_APP_SERVICES` to true.

docs/README.md
RassK: It was actually intentionally named more generally, it might need more sections about different services.

docs/azure.md
RassK: SIGNALFX_ACCESS_TOKEN and SIGNALFX_ENDPOINT_URL are the bare minimum to make it work. 
SIGNALFX_AZURE_APP_SERVICES is setup by the extension already.

Applying settings will restart the service anyway (at least it did for me).

pjanotti: > SIGNALFX_AZURE_APP_SERVICES is setup by the extension already.

ops, forgot about that...

docs/README.md
pellared: if it happens then we will change it  😉

RassK: Well, that's why the question about Webjobs also. I might have to test it then. I guess AWS would get similar layout, if I remember right, there are web and serverless components as a separate service.

docs/azure.md
RassK: > > SIGNALFX_AZURE_APP_SERVICES is setup by the extension already.
> 
> ops, forgot about that...

@pjanotti btw, just discovered that `SIGNALFX_AZURE_APP_SERVICES=1` is effectively disabling WebJobs instrumentation by mistakenly thinking it's infrastructure service.  Also non of the ENVs set by the extension are effective for WebJobs. By manually adding envs and `SIGNALFX_AZURE_APP_SERVICES=0` at least profiler attaches.

pjanotti: @RassK I don't understand why this happens, do you know the details?

RassK: @pjanotti I think it's because Webjobs are executed in Kudu scm context (has it's own w3wp process), while actual web service generates second w3wp process. 
https://github.com/signalfx/signalfx-dotnet-tracing/blob/next-ver/tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L123

image

docs/azure.md
```diff
@@ -6,5 +6,26 @@
 2. Navigate to `Development Tools` > `Extensions`.
 3. Find and install the `SignalFx .NET Tracing` extension.
 4. Navigate to `Settings` > `Configuration`.
-5. Add `New application setting` to configure receiver. See [advanced-config.md](advanced-config.md#exporter-settings) exporter settings section for more details.
+5. Add `New application setting`s to configure receiver:
```
theletterf, line 9:
```suggestion
5. Add `New application settings` to configure the receiver: 
```

theletterf (changes requested):

docs/azure.md
pjanotti: @RassK I'm not that familiar with WebJobs, according to https://docs.microsoft.com/en-us/azure/app-service/webjobs-create#acceptablefiles it seems that one has to create the job as an exe to be something that we may want to instrument. It seems a very special case, at first, my instinct tells me to keep it as it is.

RassK: @pjanotti yep, a console with generic host + webjobs sdk. We can leave it as is, just document that all env's must be set manually and make sure azure one is set off. Then it works also, just not sure if it might attach something internal additionally or not.

docs/azure.md
RassK: The button name is actually in the singular, that's why the singular form is highlighted. Maybe just to rephrase entirely differently?

pellared:
```suggestion
5. Add `New application setting` to configure the receiver: 
```

?

docs/azure.md
```diff
@@ -0,0 +1,91 @@
...
+
+    Name: SIGNALFX_ENDPOINT_URL
+    Value: (Your Collector or SignalFX ingest endpoint)
+    ```
+ 
+## Azure Functions
+
+Currently not supported. 
+
+<!-- TODO: What are the workarounds we can provide? -->
```
RassK, line 91: Anything we can suggest here? OTel.NET SDK seems to work here, but we don't have it vendored yet.

RassK: @pjanotti @pellared ⬆️

docs/azure.md
```diff
@@ -0,0 +1,91 @@
...
+    Value: (Your SIGNALFX access token)
+
+    Name: SIGNALFX_ENDPOINT_URL
+    Value: (Your Collector or SignalFX ingest endpoint)
+    ```
+    (See [advanced-config.md](advanced-config.md) for more options.)
+6. Make sure that App Service was restarted.
+
+
+## WebJobs (Experimental)
```
RassK, line 22: Marked as experimental for now... hopefully we will get some feedback once someone tries experimental stuff.

docs/azure.md
pjanotti: For AZ functions we can do https://github.com/signalfx/signalfx-dotnet-tracing#azure-function-custom-linux-docker-image

IIRC there is no way to install a profiler for Windows based functions.

docs/azure.md
pjanotti: I think it is better to record this info under internal. I'm not sure if there is demand for this capability.

docs/azure.md
```diff
@@ -18,14 +18,74 @@
...
 
-<!-- 
-## WebJobs
 
-TBD
- -->
+## WebJobs (Experimental)
 
- <!-- 
+Repeat all of the App Service section steps. Currently environment configuration must be done manually. Specially notice that `SIGNALFX_AZURE_APP_SERVICES` must be disabled. It is strongly adviced to have a separate App Service for the Webjobs.
```
pjanotti, line 24: I think that the right way to do this is to disable SIGNALFX_AZURE_APP_SERVICES for the scope of the WebJob process and call attention that if you do it for the whole box you will lose the instrumentation for the Web App.

pellared:
@RassK I propose to make separate PRs for different Azure integrations such us:
- Azure App Service
- Azure WebJobs
- Azure Functions

it will be easier to review one by one and merge.

docs/azure.md
Kielek: I agree, we have supported only App Serivices. As long as there is not such request we should put your documentation into internal-doc/jira tasks to write document for WebJobs.

It is good candidate for upstram sync phase 3.

docs/azure.md
pellared: I suggest postponing this (create a Jira story)

btw you can also find some dockerfiles here: https://github.com/signalfx/signalfx-dotnet-tracing/tree/next-ver/tracer/samples/ConsoleApp

RassK:
> @RassK I propose to make separate PRs for different Azure integrations such us:
> 
> * Azure App Service
> * Azure WebJobs
> * Azure Functions
> 
> it will be easier to review one by one and merge.

@pellared yep, seems discussions went bigger than I expected. I'll do separation.

docs/azure.md
```diff
@@ -0,0 +1,19 @@
+# Azure services instrumentation guide
```
pellared, line 1:
```suggestion
# Azure instrumentation guide
```

to match with main `README.md`

docs/azure.md
```diff
@@ -0,0 +1,19 @@
...
+
+## App Service
+
+1. Choose your App Service.
+2. Navigate to `Development Tools` > `Extensions`.
+3. Find and install the `SignalFx .NET Tracing` extension.
+4. Navigate to `Settings` > `Configuration`.
+5. Add `New application setting`s to configur the receiver:
```
pellared, line 9:
```suggestion
5. Add `New application setting`s to configure the receiver:
```

docs/azure.md
```diff
@@ -0,0 +1,19 @@
...
+
+    ```
+    Name: SIGNALFX_ACCESS_TOKEN 
+    Value: (Your SIGNALFX access token)
+
+    Name: SIGNALFX_ENDPOINT_URL
+    Value: (Your Collector or SignalFX ingest endpoint)
+    ```
+    (See [advanced-config.md](advanced-config.md) for more options.)
+6. Make sure that App Service was restarted.
```
pellared, line 19: maybe simply:

```suggestion
6. Restart the App Service.
```

pellared (approved):
few suggestions (one is a typo)

PR #301, opened 2022-01-09
Compile thread sampling on Linux

johnbley:
Get thread sampling code to compile on Linux
 
It does not actually work yet; more porting needs to be done for thread creation and performance counters, but that will be a following PR.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
@@ -497,9 +497,13 @@ int GetSamplingPeriod()
     }
     try
     {
+#ifdef _WIN32
+        int ival = std::stoi(val);
+#else
         std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t> convert;
         std::string str = convert.to_bytes(val);
         int ival = std::stoi(str);
+#endif
         return (int) fmaxl(MINIMUM_SAMPLE_PERIOD, ival);
     }
     catch (...)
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -242,7 +242,7 @@ void ThreadSamplesBuffer::writeString(WSTRING& str)
 {
     // limit strings to a max length overall; this prevents (e.g.) thread names or
     // any other miscellaneous strings that come along from blowing things out
-    short usedLen = (short) min(str.length(), MAX_STRING_LENGTH);
+    short usedLen = (short) fminl(str.length(), MAX_STRING_LENGTH);
```
Kielek, line 245: Consider 
`#include <algorithm>`
then
`#define MAX_STRING_LENGTH 512UL`
and
`hort usedLen = (short) std::min(str.length(), MAX_STRING_LENGTH);`.

It should allow us to remove compilation warnings such us for every future PR `'argument': conversion from 'unsigned __int64' to 'long double', possible loss of data.`.

Similar changes should be done for second `fminl` usage.

algorithms header is included in other cpp files.

Ref: https://stackoverflow.com/questions/16584558/the-difference-between-max-and-fmax-cross-platform-compiling

johnbley: Done!

johnbley: Also, today I learned about #NOMINMAX in windows.h.

Kielek (approved):

PR #309, opened 2022-01-12
[WIP] Add Azure Webjobs docs

RassK:
## What

Adds Azure Webjobs docs

### DIFF
--- a/docs/azure.md
+++ b/docs/azure.md
@@ -17,3 +17,69 @@
     ```
     (See [advanced-config.md](advanced-config.md) for more options.)
 6. Restart the App Service.
+
+## WebJobs (Experimental)
+
+Repeat all of the App Service section steps. Currently environment configuration must be done manually. Specially notice that `SIGNALFX_AZURE_APP_SERVICES` must be disabled. It is strongly adviced to have a separate App Service for the Webjobs.
+
+    ```
+    ## Enables .NET Framework instrumentation
+
+    Name: COR_ENABLE_PROFILING
+    Value: 1
+
+    Name: COR_PROFILER
+    Value: {B4C89B0F-9908-4F73-9F59-0D77C5A06874}
+
+    ## version must match SignalFx .NET Tracing extension version
+    Name: COR_PROFILER_PATH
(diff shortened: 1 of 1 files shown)


docs/azure.md
```diff
@@ -0,0 +1,86 @@
...
+    Name: SIGNALFX_ENDPOINT_URL
+    Value: (Your Collector or SignalFX ingest endpoint)
+    ```
+    (See [advanced-config.md](advanced-config.md) for more options.)
+6. Make sure that App Service was restarted.
+
+
+## WebJobs (Experimental)
+
+Repeat all of the App Service section steps. Currently environment configuration must be done manually. Specially notice that `SIGNALFX_AZURE_APP_SERVICES` must be disabled. It is strongly adviced to have a separate App Service for the Webjobs.
```
RassK, line 24: > I think that the right way to do this is to disable SIGNALFX_AZURE_APP_SERVICES for the scope of the WebJob process and call attention that if you do it for the whole box you will miss some of the AZ specific tags and metrics - or do we lose the instrumentation for the whole Web App?

RassK: @pjanotti let's continue here that topic.

pjanotti: @RassK I prefer to have the WebJobs under internal: there we can group Azure and link to the App Service instructions.

RassK: I'm just wondering if under internal we find enough users?

pellared: Do we want users to use it? If so then I think we should test it and not call it experimental. Otherwise, we may not even get notified by the users that something is not working.

pjanotti: Typically users will be using the docs written by the doc team, not ours directly. Typically we are going to be asked for support to one of these scenarios and will add/inform the user about these features. It works better this way so we can interact closely with the users trying the scenario.

/cc @theletterf

ivomagi:
If the support for webjobs is not added in context of a particular RFE, which it seems is not, I prefer to document and publish this feature. Whether under experimental or stable is engineering call, I would base this on test coverage. If we are confident with testing coverage, I would expose this as stable.

If we would hide the support for this we would be lacking any feedback for possibly year or more, so its better to get this feedback early. Only if there is is a particular RFE behind this, we could only expose this to particular customer and get their feedback before making the support for webjobs public.

PR #310, opened 2022-01-12
Fix graphql service name mapping

lachmatt:
## Why

GraphQL integration service name mapping should work similarly to other instrumentations.

## What

Use configured service name mapping in GraphQL instrumentation.
Annotations for modified files where missing.
Removal of redundant fallback.
Addition of CustomTestFramework for better visibility into tests (basically copied from other test project).

## Tests

Added in PR.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -42,9 +42,7 @@ namespace Datadog.Trace.Configuration
         {
             Environment = source?.GetString(ConfigurationKeys.Environment);
 
-            ServiceName = source?.GetString(ConfigurationKeys.ServiceName) ??
-                          // backwards compatibility for names used in the past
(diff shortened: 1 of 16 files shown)


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -42,9 +42,7 @@ public TracerSettings(IConfigurationSource source)
         {
             Environment = source?.GetString(ConfigurationKeys.Environment);
 
-            ServiceName = source?.GetString(ConfigurationKeys.ServiceName) ??
-                          // backwards compatibility for names used in the past
-                          source?.GetString("SIGNALFX_SERVICE_NAME");
+            ServiceName = source?.GetString(ConfigurationKeys.ServiceName);
```
lachmatt, line 45: fallback is redundant after recent key rename

pellared (approved):

PR #314, opened 2022-01-13
Manual instrumentation backwards compatibility

dszmigielski:
## Why

After the upstream rebase the manual instrumentation works differently than described in docs/manual-instrumentation.md.

## What

Bringing back the design on how manual instrumentation works from main branch

## Tests

New integration test was added, unit tests were added for new classes.
Also tested end-to-end scenario using built tracer on Windows (https://app.signalfx.com/#/apm/traces/33d094204fce91ec002268fd85a9109e).


tracer/src/Datadog.Trace/IDatadogOpenTracingTracer.cs
```diff
@@ -1,4 +1,4 @@
-﻿// <copyright file="IDatadogOpenTracingTracer.cs" company="Datadog">
+// <copyright file="IDatadogOpenTracingTracer.cs" company="Datadog">
```
pellared, line 1: `// Modified by SignalFx` comment missing 😄

tracer/src/Datadog.Trace/Tracer.cs
```diff
@@ -425,5 +427,21 @@ internal Task FlushAsync()
...
+            {
+                var signalFxOpenTracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing.OpenTracing, Version=0.2.0.0, Culture=neutral, PublicKeyToken=e43a27c2023d388a"));
+                var openTracingTracerFactoryType = signalFxOpenTracingAssembly.GetType("Datadog.Trace.OpenTracing.OpenTracingTracerFactory");
+                var methodInfo = openTracingTracerFactoryType.GetMethod("RegisterGlobalTracer");
+                object[] args = { instance };
+                methodInfo?.Invoke(obj: null, args);
+            }
+            catch (Exception ex)
+            {
+                Log.Warning(ex, "Unable to load OpenTracing helper library.");
```
pellared, line 443: This will be always reported if the client does not reference `OpenTracing` in his application.
A workaround is to add the `OpenTracing.dll` to our installers. It was an issue reported by one of our clients.

dszmigielski: Yep, is that a problem? I assume if they did not add `OpenTracing.dll` they are not using this, so it shouldn't matter for them.

pellared (changes requested):
Do not forget to test it end-to-end using the installers/packages at least for Windows and Ubuntu 😉

tracer/src/Datadog.Trace/Tracer.cs
dszmigielski: I don't mind, but I would rather just change it to the INFO and leave the message as it is.

pellared: `Unable to do something` may give a feeling that something is not fine. That is why I suggest rephrasing it.

dszmigielski: Well, it's not something I'm willing to stand for too hard, I don't mind changing it as long as others find it a good solution, too :)

dszmigielski: I have changed it to
```csharp
Log.Information(ex, "OpenTracing integration was not loaded.");
```
for now

tracer/src/Datadog.Trace/Tracer.cs
```diff
@@ -425,5 +427,21 @@ internal Task FlushAsync()
         {
             return TracerManager.AgentWriter.FlushTracesAsync();
         }
+
+        private static void RegisterGlobalTracer(Tracer instance)
+        {
+            try
+            {
+                var signalFxOpenTracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing.OpenTracing, Version=0.2.0.0, Culture=neutral, PublicKeyToken=e43a27c2023d388a"));
```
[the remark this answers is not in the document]
dszmigielski: 👍 done

tracer/src/Datadog.Trace.OpenTracing/OpenTracingScopeManager.cs
```diff
@@ -0,0 +1,30 @@
+// Modified by SignalFx
+using OpenTracing;
```
pellared, line 2: nit: can you add a blank line for consistency?

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracer.cs
```diff
@@ -18,7 +18,7 @@ internal class OpenTracingTracer : global::OpenTracing.ITracer
         private readonly Dictionary<string, ICodec> _codecs;
 
         public OpenTracingTracer(IDatadogOpenTracingTracer datadogTracer)
-            : this(datadogTracer, new global::OpenTracing.Util.AsyncLocalScopeManager())
+            : this(datadogTracer, new OpenTracingScopeManager(datadogTracer.ScopeManager))
```
pellared, line 21: can you please explain why we are changing this?

is it to support code like below?

```csharp
            var signalFxScope = _signalFxTracer.StartActive("testSpan");
            var openTracingScope = new OpenTracingScope((Scope)signalFxScope);
```

if so then do we need it? it is not documented in `manual-instrumentation.md` nor I cannot find it in https://github.com/signalfx/tracing-examples/tree/main/dotnet-manual-instrumentation

tracer/src/Datadog.Trace.OpenTracing/OpenTracingScopeManager.cs
```diff
@@ -0,0 +1,30 @@
+// Modified by SignalFx
```
Kielek, line 1:
```suggestion
// Modified by Splunk Inc.

```

tracer/src/Datadog.Trace.OpenTracing/OpenTracingScopeManager.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public global::OpenTracing.IScope Activate(global::OpenTracing.ISpan span, bool finishSpanOnDispose)
+        {
+            if (ScopeManager == null)
+            {
+                return null;
+            }
+
+            return new OpenTracingScope(ScopeManager.Activate(((OpenTracingSpan)span).Span, finishSpanOnDispose));
```
pjanotti, line 27: Do we need to worry if `span` is `null` here?

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracerFactory.cs
```diff
@@ -17,6 +18,8 @@ namespace Datadog.Trace.OpenTracing
     /// </summary>
     public static class OpenTracingTracerFactory
     {
+        private static readonly object Lock = new();
```
pjanotti, line 21: I don't think this lock is necessary: since we are dealing with the statics of OpenTracing for the registration I doubt that that code doesn't have a lock already - double check my assumption, please.

pjanotti: I took a quick look there and it seems protected, it also seems that `RegisterIfAbsent` is actually a better method to be called below, see https://github.com/opentracing/opentracing-csharp/blob/461c6edd7b4209fe353eea7b927dd7046986e2b8/src/OpenTracing/Util/GlobalTracer.cs#L75

pjanotti: Anyway, double check I looked at it very quickly.

tracer/src/Datadog.Trace.OpenTracing/OpenTracingScopeManager.cs
dszmigielski: As for how we are using it right now - nope

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracer.cs
dszmigielski: Well yeah, that's an overkill. I will revert that

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracerFactory.cs
```diff
@@ -45,6 +44,16 @@ public static class OpenTracingTracerFactory
...
         }
 
+        /// <summary>
+        /// Registers a Tracer as the global tracer, if unregistered.
+        /// </summary>
+        /// <param name="tracer">Existing SignalFx Tracer instance.</param>
+        /// <returns>True if the Tracer was successfully registered by this call, otherwise false.</returns>
+        public static bool RegisterGlobalTracer(Tracer tracer)
+        {
+            return GlobalTracer.RegisterIfAbsent(WrapTracer(tracer));
```
pellared, line 54: Why we are using `RegisterIfAbsent`? Do we want to log something if it returns `false`? Maybe we should log that it was already registered?

Otherwise, we could use `GlobalTracer.Register`

Shouldn't the method be named `RegisterGlobalTracerIfAbsent` to follow the OpenTracing convention?

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
```diff
@@ -383,5 +384,18 @@ public void Parent_OverrideDefaultServiceName_SetTag()
...
+        {
+            var globalTracerRep = GlobalTracer.Instance.ToString();
+            Assert.Contains("Datadog.Trace.OpenTracing.OpenTracingTracer", globalTracerRep);
+        }
+
+        [Fact]
+        public void RegisteredAsGlobalTracer_OnlyOnce()
+        {
+            Assert.False(OpenTracingTracerFactory.RegisterGlobalTracer(_signalFxTracer));
+        }
```
pellared, line 399: nit:
we do not have to reuse `_signalFxTracer` -> we can create any new tracer instance. I have lost a few minutes to think why we are reusing `_signalFxTracer` here and I see no reason 😉 

also I am not sure what is the value of this test

pellared (approved):
minor comments only

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
dszmigielski: I mean, the code is clearer if I don't have to create it every time. Especially that I would need some of the values needed to create one to be private fields then.

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracerFactory.cs
pellared: what do you think about adding a log line when it returns `false` and renaming the method?

tracer/test/Datadog.Trace.OpenTracing.Tests/OpenTracingTracerTests.cs
pellared: > I mean, the code is clearer if I don't have to create it every time.

I disagree b/c currently, I think that there is a reason that this tracer is reused.
You can always extract a method to create the `signalFxTracer` to a method to reduce the code duplication.

still it is not a blocker for me 😉

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracerFactory.cs
```diff
@@ -49,9 +52,15 @@ public static class OpenTracingTracerFactory
...
         /// <param name="tracer">Existing SignalFx Tracer instance.</param>
         /// <returns>True if the Tracer was successfully registered by this call, otherwise false.</returns>
-        public static bool RegisterGlobalTracer(Tracer tracer)
+        public static bool RegisterGlobalTracerIfAbsent(Tracer tracer)
         {
-            return GlobalTracer.RegisterIfAbsent(WrapTracer(tracer));
+            var registeredSuccessfully = GlobalTracer.RegisterIfAbsent(WrapTracer(tracer));
+            if (!registeredSuccessfully)
+            {
+                Log.Warning("There was an error registering the OpenTracing tracer.");
```
pellared, line 60: nit

```suggestion
              Log.Warning("An OpenTracing tracer was already registered.");
```

pellared (dismissed):

tracer/test/Datadog.Trace.TestHelpers/IisFixture.cs
```diff
@@ -41,8 +41,8 @@ public void Dispose()
         {
             if (ShutdownPath != null)
             {
-                var request = WebRequest.CreateHttp($"http://localhost:{HttpPort}{ShutdownPath}");
-                request.GetResponse().Close();
+                using var httpClient = new HttpClient();
+                var response = httpClient.GetAsync($"http://localhost:{HttpPort}{ShutdownPath}").Result;
```
pellared, line 45: I think it will make the build passing however the service will still return `500` if you debug it 😢

dszmigielski: Yep, just checking if that is the root cause of tests hanging

pellared: guess: the test is hanging b/c it waits until the service shuts down

dszmigielski: Well, probably. The dispose method is called after the test, then the call to /home/shutdown is being made, and if it gets 500 it throws an exception that is never caught, so `_iisExpress.Process.Kill()` or `_iisExpress.Process.Dispose()` are never called.

dszmigielski: I am not an IIS expert, so I'm not even sure what the `/home/shutdown` endpoint is supposed to do when called. Anyway it seems that adding `SignalFx.Tracing.OpenTracing.dll` breaks it.

dszmigielski: Yep, all checks have passed now.
@pjanotti @Kielek @RassK @lachmatt @pellared does any of you have an idea why adding a dll to the tracer could break the IIS shutdown endpoint? :)

Kielek:
v 0.1.* version we have shipped OpenTracing.dll to .NET 4. 5 GAC and tracer ourput.
v 0.2.0 version does not include this library,

PR #315, opened 2022-01-13
Add SIGNALFX_LOGS_ENDPOINT_URL env var

Kielek:
## Why

Configuration needed by thread sampling exporter and the documentation for thread sampling.

## What

`SIGNALFX_LOGS_ENDPOINT_URL` env var to configure where to send logs.

## Tests

N/A

### DIFF
--- a/docs/internal/internal-config.md
+++ b/docs/internal/internal-config.md
@@ -56,6 +56,7 @@ These settings should be never used by the users.
 
 | Environment variable | Description | Default |
 |-|-|-|
+| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where thread sampling exporters send stack traces. | `http://localhost:4318/v1/logs` |
 | `SIGNALFX_METRICS_ENDPOINT_URL` | The URL to where metric exporters send metrics. | `http://localhost:9943/v2/datapoint` |
 | `SIGNALFX_METRICS_EXPORTER` | Metrics exporter to be used. It is used to encode and dispatch metrics. Available values are: `SignalFx`, `StatsD`. | `SignalFx` |
 | `SIGNALFX_RUNTIME_METRICS_ENABLED` | Enable to activate internal runtime metrics sent to SignalFx. | `false` |
--- a/docs/internal/internal-config.md
+++ b/docs/internal/internal-config.md
@@ -56,7 +56,7 @@ These settings should be never used by the users.
 
 | Environment variable | Description | Default |
 |-|-|-|
-| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where thread sampling exporters send stack traces. | `http://localhost:4318/v1/logs` |
+| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where logs are exported using [OTLP/HTTP v1 log protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) | `http://localhost:4318/v1/logs` |
 | `SIGNALFX_METRICS_ENDPOINT_URL` | The URL to where metric exporters send metrics. | `http://localhost:9943/v2/datapoint` |
 | `SIGNALFX_METRICS_EXPORTER` | Metrics exporter to be used. It is used to encode and dispatch metrics. Available values are: `SignalFx`, `StatsD`. | `SignalFx` |
 | `SIGNALFX_RUNTIME_METRICS_ENABLED` | Enable to activate internal runtime metrics sent to SignalFx. | `false` |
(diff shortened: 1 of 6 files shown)


tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -166,6 +172,11 @@ private void ConfigureMetricsTransport(IConfigurationSource source, bool isWindo
                              // default value
                              "http://localhost:9943/v2/datapoint";
             MetricsEndpointUrl = new Uri(metricsEndpointUrl);
+
+            var logsEndpointUrl = source?.GetString(ConfigurationKeys.MetricsEndpointUrl) ??
```
lachmatt, line 176:
```suggestion
            var logsEndpointUrl = source?.GetString(ConfigurationKeys.LogsEndpointUrl) ??
```

docs/internal/internal-config.md
```diff
@@ -56,6 +56,7 @@ These settings should be never used by the users.
 
 | Environment variable | Description | Default |
 |-|-|-|
+| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where thread sampling exporters send stack traces. | `http://localhost:4318/v1/logs` |
```
pellared, line 59: nit

```suggestion
| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where logs are exported using [OTLP/HTTP v1 log protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) | `http://localhost:4318/v1/logs` |
```

I would document somewhere else that currently, the only logs that we are exporting are the stack traces. I imagine that in the future it would be good if the "logs correlation" feature could also export the logs to our backend.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
```diff
@@ -61,6 +61,12 @@ internal static partial class ConfigurationKeys
         /// <seealso cref="ExporterSettings.MetricsEndpointUrl"/>
         public const string MetricsEndpointUrl = "SIGNALFX_METRICS_ENDPOINT_URL";
 
+        /// <summary>
+        /// Configuration key for the URL where the Thread Sampler can send stack traces.
```
pellared, line 65: nit. I think that the stack traces are only one use case of logs telemetry. How about something like:

```suggestion
        /// Configuration key for the URL where the logs are exported.
        /// Currently the Thread Sampler sends stack traces as logs.
```

pellared (commented):
my comments are nit. by @lachmatt has a good one 😉

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -84,6 +84,12 @@ public ExporterSettings(IConfigurationSource source)
         /// <seealso cref="ConfigurationKeys.MetricsEndpointUrl"/>
         public Uri MetricsEndpointUrl { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Uri where the Thread Sampler can send stack traces.
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.LogsEndpointUrl"/>
```
Kielek, line 90:
```suggestion
        /// Configuration key for the URL where the logs are exported.
        /// Currently the Thread Sampler sends stack traces as logs.
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -84,6 +84,12 @@ public ExporterSettings(IConfigurationSource source)
         /// <seealso cref="ConfigurationKeys.MetricsEndpointUrl"/>
         public Uri MetricsEndpointUrl { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Uri where the Thread Sampler can send stack traces.
```
Kielek, line 88:
```suggestion
        /// Gets or sets the Uri where the logs are exported.
```

tracer/src/Datadog.Trace/Configuration/ImmutableExporterSettings.cs
```diff
@@ -60,6 +61,12 @@ public ImmutableExporterSettings(ExporterSettings settings)
         /// <seealso cref="ConfigurationKeys.MetricsEndpointUrl"/>
         public Uri MetricsEndpointUrl { get; }
 
+        /// <summary>
+        /// Gets the Uri where the Thread Sampler can send stack traces.
```
Kielek, line 65:
```suggestion
        /// Gets the Uri where the logs are exported.
```

pellared (approved):

pjanotti (approved):
LGTM

PR #318, opened 2022-01-14
Add SIGNALFX_REALM env variable

lachmatt:
## Why
Simplify endpoint configuration for traces and metrics.

## What

Addition of new SIGNALFX_REALM env variable.
Documentation to be updated with a separate PR.

## Tests

Included in PR.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
@@ -20,6 +20,13 @@ namespace Datadog.Trace.Configuration
         public const string AgentHost = "SIGNALFX_AGENT_HOST";
 
         /// <summary>
+        /// Configuration key for the SignalFx ingest realm, where the Tracer can send telemetry signals.
+        /// </summary>
+        /// <seealso cref="ExporterSettings.MetricsEndpointUrl"/>
+        /// <seealso cref="ExporterSettings.AgentUri"/>
+        public const string IngestRealm = "SIGNALFX_REALM";
+
+        /// <summary>
         /// Configuration key for the Agent port where the Tracer can send traces.
         /// Default value is 8126.
         /// </summary>
--- a/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
@@ -30,7 +30,9 @@ namespace Datadog.Trace.Configuration
         /// <summary>
         /// The default port value for dogstatsd.
         /// </summary>
-        internal const int DefaultDogstatsdPort = 8125;
+        internal const int DefaultDogstatsdPort = 9943;
+
+        private const string LocalIngestRealm = "none";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ExporterSettings"/> class with default values.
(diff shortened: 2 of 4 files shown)


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
```diff
@@ -19,6 +19,13 @@ internal static partial class ConfigurationKeys
         /// <seealso cref="ExporterSettings.AgentUri"/>
         public const string AgentHost = "SIGNALFX_AGENT_HOST";
```
lachmatt, line 20: I think it can be removed (in a separate PR)

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System.Collections.Specialized;
```
Kielek, line 1: // Modified by Splunk Inc.

or Created by Splunk?

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    [Fact]
+    public void If_there_is_no_metrics_endpoint_configuration_then_local_metrics_collector_is_used()
+    {
+        var collection = new NameValueCollection();
+        var configuration = new NameValueConfigurationSource(collection);
+
+        var settings = new ExporterSettings(configuration);
+
+        settings.MetricsEndpointUrl.Should().Be("http://localhost:9943/v2/datapoint");
```
Kielek, line 29: localhost vs 127.0.0.1? as I remember there is a functionality which changes localhost to 127.0.0.1 for spans endpoint

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    [Fact]
+    public void If_nondefault_signalfx_realm_is_configured_then_direct_ingest_endpoint_for_traces_is_used()
+    {
+        var collection = new NameValueCollection { { "SIGNALFX_REALM", "realm" } };
```
Kielek, line 35: consider test-realm as a value, to avoid suggestion, that it is the real one

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -48,8 +50,12 @@ public ExporterSettings()
         public ExporterSettings(IConfigurationSource source)
         {
             var isWindows = FrameworkDescription.Instance.OSPlatform == OSPlatform.Windows;
-            ConfigureTraceTransport(source, isWindows);
-            ConfigureMetricsTransport(source, isWindows);
+
+            var ingestRealm = source?.GetString(ConfigurationKeys.IngestRealm) ??
+                              LocalIngestRealm;
```
Kielek, line 55: Consider `null`? I am not sure if it good idea to use magic string to pass info that there is no ingest.

lachmatt: I added it to be aligned with gdi spec https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries

pellared (approved):

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
pellared: AFAIK this functionality is related to upstream configuration

1. I suggest not to use any magic like upstream as it may cause problems. One can intentionally change `hosts` file to redirect `localhost` to something different.
2. I suggest using `localhost` by default to be consistent with our other default values.

PR #333, opened 2022-01-18
Do not discourage setting SIGNALFX_SERVICE_NAME

pellared:
## Why

Mainly for sake of consistency with other languages/ecosystems.

## What

Do not discourage setting SIGNALFX_SERVICE_NAME as it is nothing bad if one uses it. We rather recommend specifying it explicitly.

### DIFF
--- a/docs/azure-instrumentation.md
+++ b/docs/azure-instrumentation.md
@@ -8,6 +8,11 @@
 4. Go to **Settings > Configuration**.
 5. Click **New application setting** to add the following settings:
 
+   Name of the instrumented service:
+   
+   * Name: `SIGNALFX_SERVICE_NAME`
+   * Value: `[my-service-name]`
+
    Access token: See [here](https://docs.splunk.com/Observability/admin/authentication-tokens/org-tokens.html) to learn how to obtain one. 
 
    * Name: `SIGNALFX_ACCESS_TOKEN`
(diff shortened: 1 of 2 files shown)


docs/azure-instrumentation.md
```diff
@@ -8,6 +8,11 @@
 4. Go to **Settings > Configuration**.
 5. Click **New application setting** to add the following settings:
 
+   Name of the instrumented service:
+   
+   * Name: `SIGNALFX_SERVICE_NAME`
+   * Value: `[my-service-name]`
```
pjanotti, line 14: Mention the default value in this case: it is listed on the advanced settings file.

pjanotti (approved):

docs/azure-instrumentation.md
pellared: 1. Do we want to mention the default value in all places? e.g. in the main `README.md` and `windows-service-instrumentation.md`?

2. Personally, if I would be the SRE, I would rather prefer to set it explicitly to make sure that e,g, it does not change in the future e.g. if the dev team would change the assembly entry name. I see the "service.name" as something like ID and I think it makes sense to recommend setting it explicitly.

pjanotti: You persuaded me, keep as it is: it is simple to setup. The only case that really gets tricky is IIS with shared app pool the others is fine to add it since it is pretty simple.

pjanotti (approved):

Kielek (approved):

PR #335, opened 2022-01-19
Add bug bash to the release process

pellared:
## Why

I find [bug bash](https://en.wikipedia.org/wiki/Bug_bash) as a very pragmatic testing tool. While automated tests can find a lot of issues their test suite is "closed". Moreover, nothing replaces the creativity of a human being trying to break things. If colleagues from other teams could be involved they can stress also other problems that we are not aware of  (like inconsistencies between languages, not clear documentation). At last bug bash is usually end-to-end testing therefore it helps finding issues in other systems such as Collector or Backend.

Side note: I remember that we had periodic bug bashes organized by @flands. I found them useful also as an onboarding experience (and I was able to find issues as a newbie),

## What

Add bug bash to the release process.

### DIFF
--- a/docs/internal/releasing.md
+++ b/docs/internal/releasing.md
@@ -14,7 +14,7 @@
    it is time to perform a [bug bash](https://en.wikipedia.org/wiki/Bug_bash).
    When testing, we should:
    
-   - use different testing environemnts,
+   - use different testing environments,
    - use different sample applications,
    - test different features,
    - follow the documentation (e.g. do not use the developer's build output,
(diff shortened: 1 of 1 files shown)


Kielek:
I do not like the idea of manual regression phase. Before release we should check the installation process for msi/other packages. Other things should be covered by our CI process.

If we speaking about cross GDI Bug Bash - I cannot imagine that we will organize for each release (not only .NET release).

pellared:
> I do not like the idea of manual regression phase. Before release we should check the installation process for msi/other packages. Other things should be covered by our CI process.

The idea is not to have a "regression test" but "exploration test".

The testing installation packages should be automated. I propose to create stories for instead of doing it manually.

pellared:
I see no interest. Closing

PR #340, opened 2022-01-20
Profiling smoke test

Kielek:
## Why

To include thread sampling in CI.

## What

Smoke test for thread sampling.
It should be modified when the exporter will be ready.

## Tests

Just a test :)

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
@@ -1,0 +1,38 @@
+// Modified by Splunk Inc.
+
+#if !NETFRAMEWORK
+
+using Datadog.Trace.TestHelpers;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests
+{
+    public class ThreadSamplingTests : TestHelper
+    {
+        public ThreadSamplingTests(ITestOutputHelper output)
+            : base("ThreadSampling", output)
+        {
+            SetServiceVersion("1.0.0");
+        }
+
+        [SkippableFact]
+        [Trait("Category", "EndToEnd")]
+        [Trait("RunOnWindows", "True")]
+        public void SubmitTheadSamples()
+        {
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_ENABLED", "true");
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_PERIOD", "1000");
+
+            using (var agent = EnvironmentHelper.GetMockAgent())
+            using (var processResult = RunSampleAndWaitForExit(agent.Port))
+            {
+                // at application works 5 seconds, we should expect at least 3 attempts of thread sampling
+                processResult.StandardOutput.Should().Contain("thread samples captured at", AtLeast.Times(expected: 3));
+            }
+        }
+    }
+}
+
+#endif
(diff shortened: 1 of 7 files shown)


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -0,0 +1,38 @@
...
+        public void SubmitTheadSamples()
+        {
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_ENABLED", "true");
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_PERIOD", "1000");
+
+            using (var agent = EnvironmentHelper.GetMockAgent())
+            using (var processResult = RunSampleAndWaitForExit(agent.Port))
+            {
+                // at application works 5 seconds, we should expect at least 3 attempts of thread sampling
+                processResult.StandardOutput.Should().Contain("thread samples captured at", AtLeast.Times(expected: 3));
```
pellared, line 32: do we have any Jira story to remove printing to stdout (and any other "functional" side-effects)?
the instrumentation should not visibly affect the application 😉 

for the beta release, I think we can also consider printing such info to stdout only if `SIGNALFX_TRACE_DEBUG=true`

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -0,0 +1,38 @@
...
+            SetServiceVersion("1.0.0");
+        }
+
+        [SkippableFact]
+        [Trait("Category", "EndToEnd")]
+        [Trait("RunOnWindows", "True")]
+        public void SubmitTheadSamples()
+        {
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_ENABLED", "true");
+            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_PERIOD", "1000");
```
pellared, line 26:
```suggestion
            SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_PERIOD", "100");
```

so that the `Thread.Sleep` in the sample app can be decreased to `0.5 sec` 

ps. the same change would need to be done in `tracer/test/test-applications/integrations/Samples.ThreadSampling/Properties/launchSettings.json`

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -0,0 +1,38 @@
+// Modified by Splunk Inc.
+
+#if !NETFRAMEWORK
```
pellared, line 3: nit: we can give a comment that Thread sampling is not supported for .NET Framework

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
Kielek: The real period is the max (SIGNALFX_THREAD_SAMPLING_PERIOD, 1000). We cannot do this :)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
Kielek: Yes, we have task to remove any logs for std. out. Extended about the additional flag.

pellared (approved):

pellared:
> ## Tests
> 
> Just a test :)

You can check if it fails if you comment this line `SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_ENABLED", "true");`  😄

Kielek:
> > ## Tests
> > Just a test :)
> 
> You can check if the test fails if you comment this line `SetEnvironmentVariable($"SIGNALFX_THREAD_SAMPLING_ENABLED", "true");` 😄

Tested, even failed on CI on .NET Framework 4.6.1 before I added preprocessor if ;)

PR #341, opened 2022-01-20
remove unused pipelines

Kielek:
## Why

Cleanup unused code to remove following actions
image

## What

Remove unused azure pipelines

## Tests

CI


pellared:
This PR removes also some GH workflows. Is it intended? When I quickly looked at them I found them interesting and maybe we can consider reusing them.

Kielek:
> This PR removes also some GH workflows. Is it intended? When I quickly looked at them I found them interesting and maybe we can consider reusing them.

PR title fixed.
As long as we do not using them it should be safe to remove it. Revert process is simple.

PR #342, opened 2022-01-20
More minor thread sampling tweaks

johnbley:
doc comments, error handling, better type alignment, etc.


tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -57,12 +57,18 @@ void ThreadSampling_RecordProducedThreadSample(std::vector<unsigned char>* buf)
     } else if (bufferB == NULL) {
         bufferB = buf;
     } else {
+        trace::Logger::Warn("Unexpected buffer drop in ThreadSampling_RecordProducedThreadSample");
         delete buf; // needs to be dropped now
     }
 }
 // Can return 0 if none are pending
-int ThreadSampling_ConsumeOneThreadSample(int len, unsigned char* buf)
+int32_t ThreadSampling_ConsumeOneThreadSample(int32_t len, unsigned char* buf)
```
pellared, line 65: can we use `size_t` instead of `int32_t` (and also get rid of `int` cast inside the definition) ?

johnbley: No, I'm trying to ensure that the types line up better with the calling code in managed-space: https://docs.microsoft.com/en-us/dotnet/standard/native-interop/type-marshaling.  `size_t` (and the previous `int` for that matter) is variable width and won't marshal properly.

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -81,7 +87,7 @@ int ThreadSampling_ConsumeOneThreadSample(int len, unsigned char* buf)
     {
         return 0;
     }
-    int toUseLen = (int) std::min(toUse->size(), (size_t) len);
+    size_t toUseLen = (int) std::min(toUse->size(), (size_t) len);
     memcpy(buf, toUse->data(), toUseLen);
     delete toUse;
     return toUseLen;
```
Kielek, line 93: It will be great if you can fix 
```
tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp#L93
'return': conversion from 'size_t' to 'int', possible loss of data
```
before merge. Otherwise, we will have eavery PR polluted by this message.

Kielek (approved):

PR #346, opened 2022-01-21
Fix OpenTracing scope

RassK:
## What

Fixes scoping if scope is opened by OpenTracing tracer.

## Tests

followed by OpenTracing Log Injection samples


pjanotti (commented):
I don't understand exactly how/why it is broken. Can you add a test showing the case that is broken?

It is also not clear to me when to use the concrete types or the interfaces instead.

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracer.cs
```diff
@@ -18,7 +18,7 @@ internal class OpenTracingTracer : global::OpenTracing.ITracer
         private readonly Dictionary<string, ICodec> _codecs;
 
         public OpenTracingTracer(IDatadogOpenTracingTracer datadogTracer)
-            : this(datadogTracer, new global::OpenTracing.Util.AsyncLocalScopeManager())
```
pjanotti, line 21: Why this doesn't work?

RassK: OT_vs_DD

The upper image is about the current logic, the bottom one is what we had in the old main branch. Simply saying that if the OT creates ScopeManager of it's own, without interfacing DD ScopeManager, then DD ScopeManager doesn't know about scopes that OT tracer opened... so log injection doesn't have trace/span data to inject.

Kielek (approved):
I agree that it would be great to avoid changing `ISpan` to `Span` in interfaces.

I think that resolving it, could be too big for current PR. Please consider follow up PR/separate jira item.

RassK:
> I agree that it would be great to avoid changing `ISpan` to `Span` in interfaces.
> 
> I think that resolving it, could be too big for current PR. Please consider follow up PR/separate jira item.

I would leave it entirely for now. If the affected scope would be only Datadog.Trace.OpenTracing, then fine let's do it. But as most of the OT stuff is just wrappers around DD tracer components, then it would include significant changes in internal types + prob needs more internal interfaces. If anyone wants to look, then the tree starts from `ScopeManagerBase.Activate(Span span, bool finishOnClose)`

tracer/src/Datadog.Trace.OpenTracing/OpenTracingTracer.cs
pjanotti: I see, I was assuming that somehow `OpenTracing.Util.AsyncLocalScopeManager()` interacted with the `GlobalTracer.Instance` to get this to work and that is not the case.

pjanotti (approved):
@RassK thanks for the explanation. Could you please create a story to ensure this doesn't regress in the future?

PR #348, opened 2022-01-21
Add tests to thread sampling batch parser

pjanotti:
## Why

1. Need to parse thread samples to a format that can be sent to the backend.
2. Need tests to validate that future changes won't break the parsing

## What

* Changed thread sample parser to generate a type, ThreadSample, that can be serialized and sent to backend;
* Added verification tests to help avoid regressions on thread sample batch parser;

## Tests

* New verification tests for the parser;
* Manually validated in a separate branch that the parsed thread samples are correctly consumed by the backend;


pjanotti:
The bin files can be reduced but as it is right now they reflect what the managed code actually receives. I will look at reducing their size in a separate PR.

johnbley (approved):
Looks good and a definite improvement from the "style" of the prototype code.

PR #349, opened 2022-01-24
enable alpine in build pipeline

Kielek:
## Why

.NET supports alpine by default.

## What

Enable build pipeline on Linux Alpine.

## Tests

CI
tar.gz layout same as for Debian.


lachmatt (approved):

theletterf:
@Kielek @lachmatt  This means that we're Alpine compatible? If that's the case, we should update the docs.

Kielek:
> @Kielek @lachmatt This means that we're Alpine compatible? If that's the case, we should update the docs.

I think that GitHub documentation will be updates in this PR. The _official_ documentation should be updated when we release next version after merging this PR (probably 0.2.1 version).

theletterf (approved):

PR #351, opened 2022-01-25
Add basic thread samples exporter

pjanotti:
## Why

To enable export thread samples easily so we can start validate and experiment with the profiling feature.

## What

* Add a basic thread sample exporter. This exporter leverage the managed thread already used to parse the thread sample batches and simply exports the last batch via OTLP logs following the GDI conventions;
* Removed all stdout logs;

## Tests

* Update smoke test for profiling to check the logs;
* Manually sent data to backend and validated the data;

### DIFF
--- a/tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
+++ b/tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
@@ -1,0 +1,173 @@
+// Modified by Splunk Inc.
+
+using System;
+using System.Collections.Generic;
+using System.Net;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Propagation;
+using Datadog.Tracer.OpenTelemetry.Proto.Common.V1;
+using Datadog.Tracer.OpenTelemetry.Proto.Logs.V1;
+using Datadog.Tracer.OpenTelemetry.Proto.Resource.V1;
+
+namespace Datadog.Trace.ThreadSampling
+{
+    internal class ThreadSampleExporter
+    {
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ThreadSampleExporter));
+
+        private readonly List<KeyValue> _fixedLogRecordAttributes;
+        private readonly Uri _logsEndpointUrl;
+        private readonly LogsData _logsData;
+
+        internal ThreadSampleExporter(ImmutableTracerSettings tracerSettings)
+        {
+            _fixedLogRecordAttributes = new List<KeyValue>
+            {
+                GdiProfilingConventions.LogRecord.Attributes.Source,
+                GdiProfilingConventions.LogRecord.Attributes.Period((long)tracerSettings.ThreadSamplingPeriod.TotalMilliseconds)
+            };
+
+            _logsEndpointUrl = tracerSettings.ExporterSettings.LogsEndpointUrl;
+
+            _logsData = GdiProfilingConventions.CreateLogsData();
+        }
+
+        public void ExportThreadSamples(List<ThreadSample> threadSamples)
+        {
+            if (threadSamples == null || threadSamples.Count < 1)
+            {
+                return;
+            }
+
+            // Populate the actual log records.
+            List<LogRecord> logRecords = _logsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs;
+            logRecords.Clear();
+
+            for (int i = 0; i < threadSamples.Count; i++)
+            {
+                var logRecord = new LogRecord
+                {
+                    Attributes =
+                    {
+                        _fixedLogRecordAttributes[0],
+                        _fixedLogRecordAttributes[1],
+                    },
+                    Body = new AnyValue { StringValue = threadSamples[i].StackTrace },
+                    TimeUnixNano = threadSamples[i].Timestamp,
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        public void ExportThreadSamples(List<ThreadSample> threadSamples)
+        {
+            if (threadSamples == null || threadSamples.Count < 1)
+            {
+                return;
+            }
+
+            // Populate the actual log records.
+            List<LogRecord> logRecords = _logsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs;
```
Kielek, line 44: I hope I understand the code correctly. If you are trying to reuse the same _logsData to avoid unnecessary allocation it will be great to have a comment for this why you  `Clear()` it.

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                Log.Error("Exception sending thread samples to {0}: {1}", _logsEndpointUrl, ex.Message);
+            }
+        }
+
+        /// <summary>
+        /// Holds the GDI profiling semantic conventions.
+        /// <see href="https://github.com/signalfx/gdi-specification/blob/b09e176ca3771c3ef19fc9d23e8722fc77a3b6e9/specification/semantic_conventions.md#profiling-resourcelogs-message"/>
+        /// </summary>
+        private static class GdiProfilingConventions
```
Kielek, line 103: What's about `SIGNALFX_TRACE_GLOBAL_TAGS`. Do you think that we should include it here?

pjanotti: Yes, this seems like a good idea.

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
```diff
@@ -0,0 +1,173 @@
...
+            httpWebRequest.Headers.Add(CommonHttpHeaderNames.TracingEnabled, "false");
+
+            using (var stream = httpWebRequest.GetRequestStream())
+            {
+                Vendors.ProtoBuf.Serializer.Serialize(stream, _logsData);
+                stream.Flush();
+            }
+
+            // Release the log records as soon as possible.
+            _logsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs.Clear();
```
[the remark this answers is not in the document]
pjanotti: Good catch, one is a leftover.

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            using (var stream = httpWebRequest.GetRequestStream())
+            {
+                Vendors.ProtoBuf.Serializer.Serialize(stream, _logsData);
+                stream.Flush();
+            }
+
+            // The exporter reuses the _logsData object, but the actual log records are not
+            // needed after serialization, release the log records so they can be garbage collected.
+            _logsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs.Clear();
```
Kielek, line 83: One more thing, is there any chance that Exception occurs between we put data to _logsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs and the Clear method? If so, we could stay with not removed data in the common collection.

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleNativeFormatParser.cs
```diff
@@ -52,8 +57,17 @@ internal List<ThreadSample> Parse()
                     }
 
                     long sampleStartMillis = ReadInt64();
-                    var sampleStart = new DateTime((sampleStartMillis * TimeSpan.TicksPerMillisecond) + TimeConstants.UnixEpochInTicks).ToLocalTime();
-                    Console.WriteLine("thread samples captured at " + sampleStart.ToLongDateString() + " " + sampleStart.ToLongTimeString());
+                    batchTimestampNanos = (ulong)sampleStartMillis * 1_000_000u;
+
+                    if (Log.IsEnabled(LogEventLevel.Debug))
```
RassK, line 62: Could be cached, no need to check in every cycle. Maybe even in field level, as we do not change it after startup.

tracer/src/Datadog.Trace/ThreadSampling/ThreadSampleExporter.cs
pjanotti: Good question: my assumption was only OOM because we are using only HttpWebRequest in synchronous mode. However, it is possible to crash here via a misconfiguration of the URL for the logs exporter: if the schema is not "http" or "https" the Uri can be created successfully but it will throw an exception every time that we try to export.

While validating this error path I noticed that many errors in the configuration are not handled. I created a [ticket](https://signalfuse.atlassian.net/browse/APMI-2693) to track the general issue with the configuration.

For this PR I will protect it with a try-catch-finally.

RassK (approved):
Looks good, just some nits to decide, not gonna be blockers.

PR #356, opened 2022-01-25
Treat warnings as errors in Linux build

johnbley:
Treat warnings as errors in Linux build (and fix the existing warnings).

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
@@ -329,7 +329,7 @@ private:
         ClassID parentClassID;
         HRESULT hr = S_OK;
 
-        if (classId == NULL)
+        if (classId == 0)
         {
             Logger::Debug("NULL classId passed to GetClassName");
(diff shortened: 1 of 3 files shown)


pjanotti (approved):
LGTM

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
```diff
@@ -329,7 +329,7 @@ class SamplingHelper
         ClassID parentClassID;
         HRESULT hr = S_OK;
 
-        if (classId == NULL)
+        if (classId == 0)
```
Kielek, line 332: I am not a C++ expert, but I have seen `nullptr` in never C++ version. Is there an option to apply it here?

Kielek (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/ThreadSampler.cpp
johnbley: Those aren't pointers, which is why it was complaining :-)

PR #358, opened 2022-01-26
Fix operation order

lachmatt:
## Why

Fix reference usage.
## What

Reorder operations in `NameCache::put`.

## Tests

Existing


Kielek (approved):

johnbley (approved):
Yep, make sense.

PR #359, opened 2022-01-26
Set warnings as errors in Windows native build

johnbley:
Mirroring what was done for Linux recently.  Currently no warnings to fix; might do the work to increase warning level from /W3 to /W4 in the future.


pjanotti:
@johnbley 
> might do the work to increase warning level from /W3 to /W4 in the future.

was it at w4 before? I think we should keep the same warning level as upstream.

pjanotti (approved):

johnbley:
> was it at w4 before? I think we should keep the same warning level as upstream.

Nope, I left it alone in this PR.

PR #360, opened 2022-01-27
Update thread sampling smoke tests

Kielek:
## Why

To have more robust tests and avoid testing stdout.

## What

Thread sampling test utilize mocked version of Open Telemetry Logs Collector Protobuf/Http endpoint.

## Tests

CI.
Manually tested, if application is not producing stack traces it fails.

## Notes

For now, it is validating only number sets of stack traces.
The plan is to introduce more in separate PR.


opentelemetry/proto/logs/v1/logs.proto
```diff
@@ -24,6 +24,7 @@ option java_package = "io.opentelemetry.proto.logs.v1";
 option java_outer_classname = "LogsProto";
 option go_package = "github.com/open-telemetry/opentelemetry-proto/gen/go/logs/v1";
 option csharp_namespace = "Datadog.Tracer.OpenTelemetry.Proto.Logs.V1";
+// option csharp_namespace = "OpenTelemetry.TestHelpers.Proto.Logs.V1";
```
pjanotti, line 27: Why this commented line? (Similar in other files).

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -26,17 +26,14 @@ public void SubmitTheadSamples()
...
             using (var agent = EnvironmentHelper.GetMockAgent())
-            using (var processResult = RunSampleAndWaitForExit(agent.Port))
+            using (var logsCollector = EnvironmentHelper.GetMockOtelLogsCollector())
+            using (var processResult = RunSampleAndWaitForExit(agent.Port, logCollectorPort: logsCollector.Port))
             {
+                var logsData = logsCollector.LogsData.ToArray();
                 // The application works for 5 seconds with debug logging enabled we expect at least 2 attempts of thread sampling in CI.
                 // On a dev box it is typical to get at least 3 but the CI machines seem slower, using 2 until the test is improved.
-                processResult.StandardOutput.Should().Contain("thread samples captured at", AtLeast.Times(expected: 2));
+                logsData.Length.Should().BeGreaterOrEqualTo(expected: 2);
```
pjanotti, line 36: A nice addition for the future is to use Verify to ensure that the stacks are looking as expected, anyway not on this PR.

pjanotti (approved):
LGTM,

opentelemetry/proto/logs/v1/logs.proto
Kielek: We have two separate (de)serializer, for:
 * sending data to the collector  in `Datadog.Tracer.OpenTelemetry.Proto.*` namespace
 * mocked collector in `OpenTelemetry.TestHelpers.Proto.*` namespace

I can make a copy of these files for testing purposes or put alternative namespace if anybody would like to regenerate it.

pjanotti: Got it, no need to separate the files just trying to understand the reason behind it.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
pjanotti: https://github.com/signalfx/signalfx-dotnet-tracing/pull/364

PR #363, opened 2022-01-28
Bump Verify.XUnit from 11.12.1 to 15.1.1

Kielek:
## Why

Version 13.2.0 brings `DisableRequireUniquePrefix` which will be useful for Thread Sampling tests.

## What

Bump Verify.XUnit from 11.12.1 to 15.1.1

## Tests

CI

PR #364, opened 2022-01-28
verify stack traces in thread sampling feature

Kielek:
## Why & What

Verify content of Sampled Stack Traces.

The application always return the same stack traces, so the verify functionality is the most reasonable way how to handle it.

## Tests

CI

# Notes
Extends #360 and #363


pjanotti (approved):
Excellent!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -32,8 +34,15 @@ public void SubmitTheadSamples()
...
                 var logsData = logsCollector.LogsData.ToArray();
                 // The application works for 5 seconds with debug logging enabled we expect at least 2 attempts of thread sampling in CI.
-                // On a dev box it is typical to get at least 3 but the CI machines seem slower, using 2 until the test is improved.
+                // On a dev box it is typical to get at least 3 but the CI machines seem slower, using 2
                 logsData.Length.Should().BeGreaterOrEqualTo(expected: 2);
+
+                var settings = VerifyHelper.GetThreadSamplingVerifierSettings();
+                foreach (var data in logsData)
+                {
+                    // all samples should be the same, as the testing application is just Thread.Sleep()ing
```
Kielek, line 43: My assumption was too optimistic.
Based on what I see in CI tests, stack traces vary on different env.
Manual verification will be needed. It is a plan on Monday.

Kielek: Fixed.
@pjanotti, please check if you are fine with current solution.

pjanotti: It seems good @Kielek

PR #365, opened 2022-01-29
Fix async profiling context

pjanotti:
## Why

The profiling context is not being correctly handled for async/await usage.

## What

* Leveraged the `valueChangedHandler` callback to correctly update the native profiler when the async context flows between threads. Added a unit test to validate the approach, plus did some manual tests and validation. However, the approach used to simplify the tests adds a cost to the handler, we will look at optimizing it when doing the [story to benchmark span context of profiling](https://signalfuse.atlassian.net/browse/APMI-2394).

## Tests

* Manually validated that the problem was real and used the same app to validate the change
* Added unit tests to ensure that it works as expected

### DIFF
--- a/tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
+++ b/tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
@@ -5,6 +5,7 @@
 
 // Modified by Splunk Inc.
 
+using System;
 using System.Threading;
 using Datadog.Trace.ClrProfiler;
 using Datadog.Trace.Logging;
@@ -15,35 +16,30 @@ namespace Datadog.Trace
     {
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(AsyncLocalScopeManager));
 
-        private readonly bool _pushScopeToNative;
-
         private readonly AsyncLocal<Scope> _activeScope = new();
 
         public AsyncLocalScopeManager(bool pushScopeToNative)
         {
-            _pushScopeToNative = pushScopeToNative;
+            _activeScope = !pushScopeToNative
+                ? new()
+                : new AsyncLocal<Scope>(
+                    (AsyncLocalValueChangedArgs<Scope> args) =>
+                    {
+                        UpdateProfilerContext(args.CurrentValue);
+                    });
         }
 
         public Scope Active
         {
             get => _activeScope.Value;
-            private set
-            {
-                _activeScope.Value = value;
-                if (_pushScopeToNative)
-                {
-                    int managedThreadId = Thread.CurrentThread.ManagedThreadId;
-                    if (value == null)
-                    {
-                        NativeMethods.SignalFxSetNativeContext(0, 0, 0, managedThreadId);
-                    }
-                    else
-                    {
-                        NativeMethods.SignalFxSetNativeContext(value.Span.TraceId.Higher, value.Span.TraceId.Lower, value.Span.SpanId, managedThreadId);
-                    }
-                }
-            }
+            private set => _activeScope.Value = value;
         }
+
+        // SetProfilingContext is a internal property to facilitate tests, since the native functions are not available
+        // on the unit tests. However, it has a cost, since it introduces indirection, per IL it causes an extra callvirt.
+        // TODO: Move this test to a location that the native API can be accessed, add a test helper on the native side
+        // and remove this delegate.
+        internal Action<ulong, ulong, ulong, int> SetProfilingContext { get; set; } = NativeMethods.SignalFxSetNativeContext;
 
         Scope IScopeRawAccess.Active
         {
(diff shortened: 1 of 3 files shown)


pjanotti:
Besides the issues with the CosmosDB emulator it seems that there is a flaky test in Kafka: https://github.com/signalfx/signalfx-dotnet-tracing/runs/5012817730?check_suite_focus=true#step:7:528

```terminal
[xUnit.net 00:03:29.53]     Datadog.Trace.ClrProfiler.IntegrationTests.KafkaTests.SubmitsTraces(packageVersion: "") [FAIL]
  Failed Datadog.Trace.ClrProfiler.IntegrationTests.KafkaTests.SubmitsTraces(packageVersion: "") [7 s]
  Error Message:
   Expected allConsumerSpans.Where(span => span.Name == $"{topic} receive") to contain 60 item(s), but found 61.

  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.TestFrameworkProvider.Throw(String message)
   at FluentAssertions.Execution.CollectingAssertionStrategy.ThrowIfAny(IDictionary`2 context)
   at Datadog.Trace.ClrProfiler.IntegrationTests.KafkaTests.SubmitsTraces(String packageVersion) in /project/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/KafkaTests.cs:line 139
 ```

File: tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs [hunk 1]
```diff
@@ -15,36 +16,31 @@ internal class AsyncLocalScopeManager : IScopeManager, IScopeRawAccess
     {
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(AsyncLocalScopeManager));
```
Kielek, line 17: Not related to changes in this PR, but the Log is not used.

tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
```diff
@@ -15,36 +16,31 @@ internal class AsyncLocalScopeManager : IScopeManager, IScopeRawAccess
...
 
-        private readonly bool _pushScopeToNative;
-
         private readonly AsyncLocal<Scope> _activeScope = new();
```
Kielek, line 19: new() can be removed. The field is now initialized by the constructor.

tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
```diff
@@ -15,36 +16,31 @@ internal class AsyncLocalScopeManager : IScopeManager, IScopeRawAccess
...
-                        NativeMethods.SignalFxSetNativeContext(value.Span.TraceId.Higher, value.Span.TraceId.Lower, value.Span.SpanId, managedThreadId);
-                    }
-                }
-            }
+            private set => _activeScope.Value = value;
         }
 
+        // SetProfilingContext is a internal property to facilitate tests, since the native functions are not available
+        // on the unit tests. However, it has a cost, since it introduces indirection, per IL it causes an extra callvirt.
+        // TODO: Move this test to a location that the native API can be accessed, add a test helper on the native side
```
Kielek, line 40: Do you have a plan to fix it in follow up or create the jira item?

tracer/test/Datadog.Trace.Tests/ThreadSampling/AsyncLocalScopeManagerTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            testThread.Join();
+
+            void AssertProfilingContextMatchesScope(Scope scope, int managedThreadId)
+            {
+                var scopeCtx = scope.Span.Context;
+                var profilingCtx = tid2ProfilingCtx[managedThreadId];
+
+                Assert.Equal(scopeCtx.TraceId.Higher, profilingCtx.TraceIdHigher);
+                Assert.Equal(scopeCtx.TraceId.Lower, profilingCtx.TraceIdLower);
+                Assert.Equal(scopeCtx.SpanId, profilingCtx.SpanId);
```
Kielek, line 83: I have tried to play a bit with the test locally.
When I have added Assert.False(true); to this method, it still passes.

It fails correctly if I put such assertion out of the new Thread scope.

Kielek (changes requested):

tracer/test/Datadog.Trace.Tests/ThreadSampling/AsyncLocalScopeManagerTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                    await Task.Delay(100);
+                    await blockingThreadTask;
+                    currentManagedThreadId = Thread.CurrentThread.ManagedThreadId;
+                }
+
+                // Context must have migrated to the new thread.
+                AssertProfilingContextMatchesScope(activeScope, currentManagedThreadId);
+
+                // Context must have been cleaned up from old thread.
+                Assert.False(tid2ProfilingCtx.ContainsKey(initialManagedThreadId));
```
Kielek, line 66: I really like the FluentAssertion, mostly due to better messages.
consider tid2ProfilingCtx.Should().ContainKey(initialManagedThreadId); and similar for others.

What is more, you can use also https://fluentassertions.com/introduction#assertion-scopes to make sure that all issues will be listed in the test results, not only the first one.

tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs
pjanotti: The plan is to use the story linked on the *What* section of the PR description - I updated the description to ensure that this is included.

tracer/test/Datadog.Trace.Tests/ThreadSampling/AsyncLocalScopeManagerTests.cs
pjanotti: ugh, while trying to make the test reliable by adding a thread I forgot that the exception won't be captured by the test runner... the code need to flow that error outside...

Kielek (approved):

PR #367, opened 2022-01-31
Validate env input values

RassK:
## Why

Some values need extra careful validation or parsing (eg string to Uri). 
If parsing error happened, application could enter to parse -> error -> log loop and cause instrumented app to slow down.
If validation error happened, it was just overwriting with default value, no logging. 

## What

Adds safe parsing and logs out if something was force changed in user configuration.
Adds minimal modifications to existing solution (extending).

## Tests

`SafeReadHelperTests.cs `

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
@@ -154,7 +154,7 @@ namespace Datadog.Trace.Configuration
         {
             return IsDefaultIngestRealm(ingestRealm) ?
                        // local collector
-                       new Uri("http://127.0.0.1:9943/v2/datapoint") :
+                       new Uri("http://localhost:9943/v2/datapoint") :
                        // direct ingest
                        new Uri($"https://ingest.{ingestRealm}.signalfx.com/v2/datapoint");
         }
@@ -163,7 +163,7 @@ namespace Datadog.Trace.Configuration
         {
             return IsDefaultIngestRealm(ingestRealm) ?
                        // local collector
-                       new Uri($"http://127.0.0.1:{agentPort}/api/v2/spans", UriKind.Absolute) :
+                       new Uri($"http://localhost:{agentPort}/api/v2/spans", UriKind.Absolute) :
                        // direct ingest
                        new Uri($"https://ingest.{ingestRealm}.signalfx.com/v2/trace", UriKind.Absolute);
         }
@@ -206,7 +206,7 @@ namespace Datadog.Trace.Configuration
         {
             var logsEndpointUrl = source?.GetString(ConfigurationKeys.LogsEndpointUrl) ??
                                   // default value
-                                  "http://127.0.0.1:4318/v1/logs";
+                                  "http://localhost:4318/v1/logs";
             LogsEndpointUrl = new Uri(logsEndpointUrl);
         }
 
--- a/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
@@ -6,6 +6,7 @@
 // Modified by Splunk Inc.
 
 using System;
+using Datadog.Trace.Configuration.Helpers;
 using Datadog.Trace.Vendors.StatsdClient.Transport;
 
 namespace Datadog.Trace.Configuration
@@ -62,12 +63,10 @@ namespace Datadog.Trace.Configuration
                 // default value
                 ?? false;
 
-            var partialFlushMinSpans = source?.GetInt32(ConfigurationKeys.PartialFlushMinSpans);
-
-            if ((partialFlushMinSpans ?? 0) <= 0)
-            {
-                PartialFlushMinSpans = 500;
-            }
+            PartialFlushMinSpans = source.SafeReadInt32(
+                key: ConfigurationKeys.PartialFlushMinSpans,
+                defaultTo: 500,
+                validators: (val) => val > 0);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


tracer/test/Datadog.Trace.Tests/Configuration/SafeReadHelperTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            Uri value = source.SafeReadUri(settingName, new Uri("http://www.temp.org", UriKind.Absolute));
+
+            Assert.Equal(new Uri(expected, UriKind.Absolute), value);
+            Assert.Equal(log, waslogged);
+        }
+
+        [Theory]
+        [InlineData("SIGNALFX_FOO_VALUE", "1", 1, 0, false)]
+        [InlineData("SIGNALFX_FOO_VALUE", "-1", 1, 0, true)]
+        [InlineData("SIGNALFX_FOO_VALUE", "abc", 1, 0, false)]
```
RassK, line 42: Missing logging comes from Configuration handling. One way is to add logging to string -> int parsing, if it fails.

tracer/test/Datadog.Trace.Tests/Configuration/SafeReadHelperTests.cs
```diff
@@ -0,0 +1,87 @@
...
+            mock.Setup(x => x.Error(It.IsAny<string>(), It.IsAny<int>(), It.IsAny<string>()))
+                .Callback<string, int, string>((message, line, file) =>
+                {
+                    if (message.Contains(key))
+                    {
+                        wasCalled(true);
+                    }
+                });
+
+            log.SetValue(null, mock.Object);
```
RassK, line 84: Any ideas making it work on  .NET Core?

Kielek: I think that, it is by design. See https://github.com/dotnet/runtime/pull/37849

You should change design a bit to have possibility to inject the log outside the class. Is this helper class have to be static?

Kielek:
> Any ideas making it work on .NET Core?

While checking it, I have found few typos. Pushed commit.

tracer/test/Datadog.Trace.Tests/Configuration/SafeReadHelperTests.cs
RassK: Yep, found out too it was a bug in early .NET FX that was never fixed until Core.
Maybe it's cleaner if we exploit it. Eg excluding logging assertions in .NET Core. Anything else could cause bigger issues, as there is no DI, excessive use of statics, etc. Prob that's why I didn't find any existing log tests also.

Kielek (approved):

tracer/src/Datadog.Trace/Configuration/Helpers/SafeReadHelper.cs
```diff
@@ -0,0 +1,77 @@
+using System;
```
pjanotti, line 1: nit: we are adding the "modified" comment to added sources.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -151,22 +150,22 @@ public int PartialFlushMinSpans
         /// </summary>
         internal TransportType MetricsTransport { get; set; }
 
-        private static string GetConfiguredMetricsEndpoint(string ingestRealm)
+        private static Uri GetConfiguredMetricsEndpoint(string ingestRealm)
         {
             return IsDefaultIngestRealm(ingestRealm) ?
                        // local collector
-                       "http://localhost:9943/v2/datapoint" :
+                       new Uri("http://127.0.0.1:9943/v2/datapoint") :
```
pjanotti, line 157: I doubt that this will cause any problem but this is a slight change in behavior: one can bind only to the IP v6 address. This saves the DNS lookup (even if only local), however, it may miss if localhost was bound to something else. I'm on the fence about this...

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -219,17 +218,17 @@ private void ConfigureTraceTransport(IConfigurationSource source, bool isWindows
...
+                defaultTo: GetConfiguredTracesEndpoint(ingestRealm, agentPort));
 
             if (string.Equals(AgentUri.Host, "localhost", StringComparison.OrdinalIgnoreCase))
             {
                 // Replace localhost with 127.0.0.1 to avoid DNS resolution.
                 // When ipv6 is enabled, localhost is first resolved to ::1, which fails
                 // because the trace agent is only bound to ipv4.
                 // This causes delays when sending traces.
-                var builder = new UriBuilder(agentUri) { Host = "127.0.0.1" };
+                var builder = new UriBuilder(AgentUri.AbsoluteUri) { Host = "127.0.0.1" };
```
pjanotti, line 231: Ah this applies to the DD agent, and should be the typical configuration for the collector but IFAIK the collector can be bound to IPV6 only via configuration.

pjanotti (approved):
LGTM @RassK - I'm just on the fence regarding changing localhost to its IP v4 address: perhaps it fits better in a separate PR so we can properly think about the effects of that.

RassK:
> LGTM @RassK - I'm just on the fence regarding changing localhost to its IP v4 address: perhaps it fits better in a separate PR so we can properly think about the effects of that.

Yep, you are right. Sounds like a separate PR.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
RassK: If collector is binding IPv6 also, then we should remove it. + currently we have mixed configurations to the same collector, as traces endpoint is always rewritten to `127.0.0.1` everything else is localhost. So traces will always use IPv4 while others could be any.

pjanotti (approved):

PR #374, opened 2022-02-07
Metric exporter retries

lachmatt:
## Why

Export metrics in presence of transient failures.
## What

Try 3 times before dropping metrics.

## Tests

Smoke tests.


tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -38,45 +62,69 @@ public void Send(DataPointUploadMessage msg)
...
                     Log.Debug($"Sent {msg.datapoints.Count} metric data points to: {_metricsEndpointAddress}.");
+                    return;
                 }
-                else
+                catch (Exception e)
                 {
-                    Log.Error($"HTTP error sending metric data to {_metricsEndpointAddress}: {webResponse.StatusCode}.");
+                    if (attemptNumber >= MaxAttempts || !IsTransient(e))
+                    {
+                        Log.Error(e, $"Dropping metrics after {attemptNumber} unsuccessful attempts sending to: {_metricsEndpointAddress}.");
```
pellared, line 98: nit: `attempts` -> `attempt(s)`

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -38,45 +62,69 @@ public void Send(DataPointUploadMessage msg)
...
+                        return;
```
pellared, line 99: Should we aggregate counter metrics to avoid data loss?

PS. For sure it is not something that we should address in this PR. Just a remark.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -14,10 +16,31 @@ namespace Datadog.Trace.SignalFx.Metrics
...
+        });
+
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(SignalFxMetricExporter));
+
         private readonly Uri _metricsEndpointAddress;
+
         private readonly string _apiToken;
+
         private readonly int _webRequestTimeoutMs;
+        private readonly TimeSpan _waitBeforeRetries;
```
pellared, line 43: These values + the number of retries could be configurable

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -14,10 +16,31 @@ namespace Datadog.Trace.SignalFx.Metrics
     /// </summary>
     internal class SignalFxMetricExporter : ISignalFxMetricExporter
     {
+        private const int MaxAttempts = 3;
+
+        private static readonly HashSet<HttpStatusCode> TransientErrorStatusCodes = new HashSet<HttpStatusCode>(new[]
+        {
+            HttpStatusCode.RequestTimeout,
+#if NETCOREAPP
+            HttpStatusCode.TooManyRequests
```
pjanotti, line 25: We don't have the constant for .NET Framework, but aren't we treating the response differently? I mean the server can still return the same 429.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -38,45 +62,69 @@ public void Send(DataPointUploadMessage msg)
...
+                    Log.Debug(e, "Transient exception encountered. Retrying sending metric data.");
                 }
-            }
-            catch (Exception ex)
-            {
-                Log.Error(ex, $"Metric data could not be sent to {_metricsEndpointAddress}.");
+
+                attemptNumber++;
+                // short wait before retrying
+                Thread.Sleep(_waitBeforeRetries);
```
pjanotti, line 107: Are we in a dedicated thread?

pjanotti (commented):
The main question for me is the different behavior for 429 between .NET Fx and Core. It seems to me that we should behave the same.

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
pellared: Yes

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -29,6 +52,7 @@ public SignalFxMetricExporter(Uri metricsEndpointAddress, string apiToken)
             _metricsEndpointAddress = metricsEndpointAddress;
             _apiToken = apiToken;
             _webRequestTimeoutMs = 500;
```
lachmatt, line 54: Let me know if you think I should increase it, I think java uses 2s

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -116,15 +116,19 @@ private static bool IsTransient(Exception ex)
...
-                return TransientErrorStatusCodes.Contains(response.StatusCode);
+                return IndicatesTransientError((int)response.StatusCode);
             }
 
             return false;
         }
+
+        private static bool IndicatesTransientError(int responseStatusCode)
+        {
+            return TransientClientErrorStatusCodes.Contains(responseStatusCode) || (responseStatusCode >= 500 && responseStatusCode <= 599);
```
pellared, line 131: how about removing the `TransientClientErrorStatusCodes` to reduce the memory allocations? also we would have all status codes in one place (higher cohesion). e.g. something like 

```suggestion
        {
            // following HTTP status codes are considered as a transient error:
            // 408 Request Timeout
            // 429 Too Many Requests
            // 5xx Server Error
            return responseStatusCode == 408 || responseStatusCode == 429 || (responseStatusCode >= 500 && responseStatusCode <= 599);
```

pellared (approved):

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
pellared: I suggest using the same values as Java

tracer/src/Datadog.Trace/SignalFx/Metrics/SignalFxMetricExporter.cs
```diff
@@ -18,12 +18,12 @@ internal class SignalFxMetricExporter : ISignalFxMetricExporter
...
 
-        private static readonly HashSet<HttpStatusCode> TransientErrorStatusCodes = new HashSet<HttpStatusCode>(new[]
+        private static readonly HashSet<int> TransientClientErrorStatusCodes = new HashSet<int>(new[]
         {
-            HttpStatusCode.RequestTimeout,
-#if NETCOREAPP
-            HttpStatusCode.TooManyRequests
-#endif
+            // RequestTimeout
+            408,
```
Kielek, line 24: I do not have strong opinion, but you can consider casting (int) HttpStatusCode.RequestTimeout. It should work.

Kielek (approved):

PR #375, opened 2022-02-08
Use "." on native to separate method from namespace

pjanotti:
## Why

Native is using "::" as the separator between namespace and method. That is not the standard on dotnet and the managed side ends up allocating a new string just to replace it.

## What

Changed the string used in native to match the dotnet pattern for names and removed the replacing on the managed side.

## Tests

Manual validation that the stacks are still being produced and look as before.

Update the binary buffers to new buffers generated by the native side already using "." as the separator.


Kielek:
Good catch!

Kielek (approved):

PR #376, opened 2022-02-08
stabilize thread sampling test

Kielek:
## Why

Make the test more stable

## What

Thread sampling test. Based on what I see in the logs. The expected stack trace sometimes is not available for the first logs data. I expect that similar scenario can occur for the last logs (Tread.Sleep finished).

Extending Thread sleep timeout to 6 seconds to make at least 3 data logs.

## Tests

CI

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
@@ -59,7 +59,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
                     {
                         if (index == 0 || index == logsData.Length - 1)
                         {
-                            // skip verification for the first and the last logs. Depends on env. the expected method can not start yest or is in a finished state
+                            // skip verification for the first and the last logs. Depending on env., the expected method may not have started or is already in a finished state
                             logRecords.Should().ContainSingle(x => ContainStackTraceForClassHierarchy(x));
                         }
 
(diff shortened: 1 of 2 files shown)


tracer/test/test-applications/integrations/Samples.ThreadSampling/Program.cs
```diff
@@ -29,7 +29,7 @@ internal static class ClassD
     public static void MethodD()
     {
         Console.WriteLine("Thread.Sleep - starting");
-        Thread.Sleep(TimeSpan.FromSeconds(value: 5));
+        Thread.Sleep(TimeSpan.FromSeconds(value: 6));
```
pellared, line 32: nit 😄 

```suggestion
        Thread.Sleep(TimeSpan.FromSeconds(6));
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ThreadSamplingTests.cs
```diff
@@ -41,22 +41,28 @@ public async void SubmitThreadSamples()
...
                     var logRecords = data.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs;
 
-                    await DumpLogRecords(data);
-
                     using (new AssertionScope())
                     {
-                        logRecords.Should().ContainSingle(x => ContainStackTraceForClassHierarchy(x));
+                        if (index == 0 || index == logsData.Length - 1)
+                        {
+                            // skip verification for the first and the last logs. Depends on env. the expected method can not start yest or is in a finished state
```
pellared, line 62:
```suggestion
                            // skip verification for the first and the last logs. Depending on env., the expected method may not have started or is already in a finished state
```

pellared (approved):

PR #386, opened 2022-02-11
Configure dependabot

pellared:
## Why

In order to maintain and control the build dependencies

## What

- Update the dependabot config to bump GitHub Actions
- Remove unused GitHub workflows
-  Update the dependabot config to bump the Dockerfile base images used by CI build

## Tests

Testing in production 😄


.github/dependabot.yml
```diff
@@ -1,5 +1,21 @@
...
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "friday"
+
+  # Maintain dependencies for Dockerfiles used in build pipeline
+  - package-ecosystem: "docker"
+    directory: "/tracer/build/_build/docker"
```
RassK, line 12: There is a one more for tests: tests/test-applications/regressions/AutomapperTests, I guess tests/test-applications/ is good enough.

.github/workflows/verify_source_generated_changes_are_persisted.yml
```diff
@@ -1,34 +0,0 @@
-name: Verify source generator changes have been persisted
```
Kielek, line 1: Why? I think it is important to execute this task on PR/on main If it is working..

pellared: You are right. I will revert it.

.github/dependabot.yml
pellared: I am not sure if we should bump these. They may be old versions as these may be regression tests for some concrete versions of .NET. I am more concerned about the "build" dependencies than the dependencies used just during tests. WDYT?

.github/workflows/verify_source_generated_changes_are_persisted.yml
pellared: Addressed 5451c3a1

.github/dependabot.yml
pellared: I see that there is no automation or real test for `AutomapperTests` (as well for most of the `tests/test-applications/regressions`

RassK: I think you are right, that makes the bot spam about the versions that actually have to be different.

RassK (approved):
LGTM

Kielek (approved):

PR #390, opened 2022-02-14
Fix profiling configuration

Kielek:
## Why

To be aligned with GDI specification.

## What

Change environmental variable names to be aligned with GDI Specification. The prefixes are different `SIGNALFX_` vs `SPLUNK_`
Add experimental flag for always on profiler.
Change namespace to AlwaysOnProfiler for ThreadSampling.

## Tests

CI


docs/internal/internal-config.md
```diff
@@ -22,6 +22,7 @@ These settings should be never used by the users.
 | `SIGNALFX_LOGS_DIRECT_SUBMISSION_MAX_BATCH_SIZE` | Configuration key for the maximum number of logs to send at one time  | `1000` |
 | `SIGNALFX_LOGS_DIRECT_SUBMISSION_MAX_QUEUE_SIZE` | Configuration key for the maximum number of logs to hold in internal queue at any one time | `100000` |
 | `SIGNALFX_LOGS_DIRECT_SUBMISSION_BATCH_PERIOD_SECONDS` | Configuration key for the time in seconds to wait between checking for batches | `2` |
+| `SIGNALFX_PROFILER_EXPERIMENTAL_ENABLED` | Enable to activate experimental features for thread sampling. | `false` |
```
pellared, line 25: is it actually used anywhere?

Kielek: It is not, but as designed in our jira task.
@pjanotti, do you really think that we should include it right now?

pellared (approved):
The only questionable thing is if we need `SIGNALFX_PROFILER_EXPERIMENTAL_ENABLED`.
Maybe create a separate PR for it so that we can easily revert this config if it occurs to not be needed?

docs/internal/internal-config.md
```diff
@@ -64,8 +65,8 @@ These settings should be never used by the users.
 
 | Environment variable | Description | Default |
 |-|-|-|
-| `SIGNALFX_LOGS_ENDPOINT_URL` | The URL to where logs are exported using [OTLP/HTTP v1 log protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) | `http://localhost:4318/v1/logs` |
+| `SIGNALFX_PROFILER_LOGS_ENDPOINT` | The URL to where logs are exported using [OTLP/HTTP v1 log protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) | `http://localhost:4318/v1/logs` |
```
RassK, line 68: `SIGNALFX_PROFILER_LOGS_ENDPOINT` Shouldn't this be just `SIGNALFX_PROFILER_ENDPOINT`, it just happens currently that they are hacked through logs.

pellared: I think it is correct. See GDI spec:

image

Kielek: @RassK, I agree with you, but it is against GDI spec.

docs/internal/internal-config.md
pjanotti: We can introduce this later as needed.

Kielek: I will split the jira item, and remove it from here

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -359,6 +359,9 @@ private static async Task WriteDiagnosticLog(TracerManager instance)
                     writer.WritePropertyName("thread_sampling_period");
                     writer.WriteValue(instanceSettings.ThreadSamplingPeriod);
 
+                    writer.WritePropertyName("thread_sampling_experimental_enabled");
```
pjanotti, line 362: Do not forget to remove this one for the experimental setting.

pjanotti (approved):

tracer/src/Datadog.Trace/TracerManager.cs
Kielek: Sure, I have just removed one commit from the PR.

PR #391, opened 2022-02-14
Document SIGNALFX_DOTNET_TRACER_HOME

pellared:


Kielek (approved):

docs/advanced-config.md
```diff
@@ -53,6 +53,7 @@ The following settings are common to most instrumentation scenarios:
 | Setting | Description | Default |
 |-|-|-|
 | `SIGNALFX_AZURE_APP_SERVICES` | Set to indicate that the profiler is running in the context of Azure App Services. | `false` |
+| `SIGNALFX_DOTNET_TRACER_HOME` | Installation location. | _Automatically set ONLY by the Windows installer_ |
```
theletterf, line 56:
```suggestion
| `SIGNALFX_DOTNET_TRACER_HOME` | Defines the location of the .NET tracer. Set by the Windows installer automatically. Must be set manually to `/opt/signalfx` when instrumenting applications on Linux or background services in Azure App Service.  |
```

theletterf (changes requested):

docs/advanced-config.md
pellared:
```suggestion
| `SIGNALFX_DOTNET_TRACER_HOME` | Installation location. Must be set manually to `/opt/signalfx` when instrumenting applications on Linux or background services in Azure App Service.  |  _Automatically set ONLY by the Windows installer_ |
```

1. I do not like the term ".NET tracer"I think we should call it ".NET auto-instrumentation"
2. Not sure if " Defines the location " it is the place where it searches and where it should be located
3. Missing column for default value 😄

theletterf (changes requested):

docs/advanced-config.md
RassK: in x86 > 
`C:\Program Files (x86)\SignalFX\.NET Tracing`

in x64 >
`C:\Program Files (x86)\SignalFX\.NET Tracing`
`C:\Program Files\SignalFX\.NET Tracing`

pellared: Commit pushed https://github.com/signalfx/signalfx-dotnet-tracing/pull/391/commits/9165da20e98213cd00294b30377a704a305206b0

theletterf (approved):

PR #392, opened 2022-02-15
Adjust naming convention

Kielek:
## Why

Cleanup POC.

## What

Adjust C# naming convention to the common standard together with some semi-auto r# fixes.
NULL->nullptr where possible for thread sampler.

## Tests

CI


tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleNativeFormatParser.cs
```diff
@@ -17,22 +17,22 @@ internal class ThreadSampleNativeFormatParser
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<ThreadSampleNativeFormatParser>();
         private static readonly bool IsLogLevelDebugEnabled = Log.IsEnabled(LogEventLevel.Debug);
 
-        private readonly byte[] buf;
-        private readonly int len;
-        private readonly Dictionary<int, string> codes;
-        private int pos;
```
dszmigielski, line 23: <3

dszmigielski (approved):

PR #393, opened 2022-02-15
Lock all dependency versions

RassK:
## Why

According to the security doc, all dependencies must be locked to specific version

## What

Locking:
* Github Actions base image versions
* APT repo packages for Debian
* (TODO) APK repo package for Alpine - Currently unsure what the final solution will be. Still under discussions in security doc.

### DIFF
--- a/tracer/build/_build/docker/debian.dockerfile
+++ b/tracer/build/_build/docker/debian.dockerfile
@@ -1,5 +1,5 @@
 # We used a fixed, older version of debian for linking reasons
-FROM mcr.microsoft.com/dotnet/runtime-deps:5.0-buster-slim as base
+FROM mcr.microsoft.com/dotnet/runtime-deps:5.0.14-buster-slim as base
 ARG DOTNETSDK_VERSION
 
 # Based on https://github.com/dotnet/dotnet-docker/blob/34c81d5f9c8d56b36cc89da61702ccecbf00f249/src/sdk/6.0/bullseye-slim/amd64/Dockerfile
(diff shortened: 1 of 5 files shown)


tracer/build/_build/docker/debian.dockerfile
```diff
@@ -19,21 +19,21 @@ ENV \
...
-        llvm \
-        clang \
-        gcc \
-        build-essential \
-        rpm \
-        ruby \
-        ruby-dev \
-        rubygems \
-    && gem install --no-document fpm \
+        git=1:2.30.2-1 \
```
pellared, line 22: shouldn't we pin the base image to `mcr.microsoft.com/dotnet/runtime-deps:5.0.14-buster-slim` ?

RassK (commented):
Github Actions configuration contains `actions/setup-dotnet@v1` step that is configured like:
```yml
with:
   dotnet-version: | 
      2.1.x
      3.1.x
      5.0.x
      6.0.200
```

... technically speaking 2.1.x, 3.1.x, 5.0.x should be more explicitly locked.  I'm not sure if there is easy way to bump these automatically.

pellared (approved):
Can you leave what is left to do?
As far as I see .NET SDK and GitHub Actions versions in our workflows.
For me it is fine to do in a separate PR (just remove `all` from the PR title 😄 ).

pellared:
> Github Actions configuration contains `actions/setup-dotnet@v1` step that is configured like:
> 
> ```yaml
> with:
>    dotnet-version: | 
>       2.1.x
>       3.1.x
>       5.0.x
>       6.0.200
> ```
> 
> ... technically speaking 2.1.x, 3.1.x, 5.0.x should be more explicitly locked. I'm not sure if there is easy way to bump these automatically.

We are already pinning `6.0.200` so we can manually bump the `3.1.x`, `5.0.x` together. BTW why do we have `2.1.x`? Now I am confused if we support it or not really.

pjanotti:
> BTW why do we have 2.1.x? Now I am confused if we support it or not really.

We don't list 2.1 but upstream supports it and we never removed build and projects.

Kielek (approved):

PR #400, opened 2022-02-21
Upstream pull

lachmatt:
## Why

Keep in sync with upstream

## What

Upstream sync 0875542586063f88ed1da6d3fe875d8b396b7a3c..6989e88ce638ba1e4ee1b178d70e692a3a5499dd

## Tests

CI


Kielek:
Code review notes, please check which of them must be fixed and which could be postponed (covered by separate jira items)

* .github\workflows\auto_create_version_bump_pr.yml should be removed
* .github\workflows\create_hotfix_branch.yml should be removed
* go through new .md files, verify content, epecially README.md's
* revert changes in shared\src\msi-installer\shared\EnvironmentVariables.wxs - ID can be called DD_, the name is correct
* shared\src\msi-installer\Tracer\EnvironmentVariables.wxs same as previous comment
* tracer\src\Datadog.Trace\Agent\TracesTransportType.cs - in previous upstream sync DAwid decided to fully remove  UnixDomainSocket things
* remove all places where SIGNALFX_APPSEC_BLOCKING_ENABLED is used. it was removed from tracer\src\Datadog.Trace\Configuration\ConfigurationKeys.cs
* Extend internal docs how to enable DD profiler and DD telemetry, AWS lambdas
* document SIGNALFX_AAS_ENABLE_CUSTOM_TRACING and SIGNALFX_AAS_ENABLE_CUSTOM_METRICS
* full DD proofiler removal(?)

Kielek:
Real environment issue
Issue summary:
Description environment: Node.js service (ardent) calls ASP.NET Core 2.x service hosted on Kestrel and targeted .NET Framework 4.6.1 (it is named HQUnchained).
Node.js  service (ardent) is instrumented by splunk-otel-js and correctly configured with b3multi propagator.
ASP.NET Core 2.x service (HQUnchained) is instrumented by the SignalFx Instrumentation for .NET v 0.2.0. Propagator is B3 (which is equivalent of b3multi on OTel side).
Current results:
Node.js service (ardent) produces Traces and span. Correct b3multi headers are send to the ASP.NET Core 2.x service (HQUnchained).
ASP.NET Core 2.x service (HQUnchained) produces only span for the db access which is not related to the trace created by the Node.js service (ardent).
Expected results:
Node.js service (ardent) produces Traces and span. Correct b3multi headers are send to the ASP.NET Core 2.x service (HQUnchained).
ASP.NET Core 2.x service (HQUnchained) produces
 - span related to the HTTP request - server (child of the span related to the HTTP request from the Node.js - client)
 - span related to the db access (child of the span related to the HTTP request - server).
 
Explanation of current results:
ASP.NET Core 2.x service (HQUnchained) is targeted for .NET Framework 4.6.1.
Instrumentation for ASP.NET Core is supported only for .NET Core 3.1 or .NET 5.0 and higher.
You can check the code for the 0.2.x version https://github.com/signalfx/signalfx-dotnet-tracing/blob/e1d2c7b9cc2a52b69b4b07644710a3a0f6e7bced/tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs#L8
I have tested the same scenario also for 0.1.6, the behavior is very similar https://github.com/signalfx/signalfx-dotnet-tracing/blob/31079b0c07e7cfc2e0b9ad24039912a80fe5ec8f/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs#L2
If the instrumentation for ASP.NET Core is not working then there is no possibility to correlate the Node.js http-client span with db-access span.
Potential workarounds:
Upgrade ASP.NET Core 2.x service (HQUnchained) to .NET (Core). Probably the best option is to target .NET 6.0 as it is LTS support. What is more .NET Framework 4.6.1 will be EOS within 2 months https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/
I do not see any other explanation that it was working for the POC. Probably POC app was written in .NET (Core) instead of the .NET Framework.
---------------
Sample app issue
Issue summary:
Sample app https://github.com/dodegaard/dotnet-otel-tester make two calls
 - Githup - http
 - postgress db
Current results:
Application produces two independent traces:
 - related to the GitHub HTTP request
 - related to the Postgres query.
 
Expected results:
Application produces one trace. 
Span related to the Postgres query is child of the GitHub HTTP request.
Explanation of current results:
The application is written in that way that the requests are fully not related.
It is oversimplified (pseudocode), but you can imagine that the auto instrumentation replaces
https://github.com/dodegaard/dotnet-otel-tester/blob/ba3067cf5c44ad34ba3fa84bc7dec344e6e8bb83/OTelDev/Program.cs#L68 by
```
StartSpan()
var response = client.GetAsync(restQuery).Result;
StopSpan()
```
and 
https://github.com/dodegaard/dotnet-otel-tester/blob/ba3067cf5c44ad34ba3fa84bc7dec344e6e8bb83/OTelDev/Program.cs#L45 by
```
StartSpan()
var output = connection.Execute(sql, null, null, null, System.Data.CommandType.Text);
StopSpan()
```
So there is not possibility to have correlation between this two spans.
I suppose the the customer would like to test code similar to following pseudocode exposed by the HTTP endpoint
```
using Microsoft.AspNetCore.Mvc;
namespace Samples.AspNetCoreMvc.Controllers
{
    public class HomeController : Controller
    {
        public IActionResult Index()
        {
            return CallDatabase();
        }
        
        private static void CallDatabase()
        {
            string sql = "select table_catalog, table_schema, table_name from information_schema.tables;";
            using (var connection = new NpgsqlConnection(connString))
            {
                try
                {
                    var output = connection.Execute(sql, null, null, null, System.Data.CommandType.Text);
                }
                catch (Exception e)
                {
                    Console.WriteLine(e.ToString());
                    throw;
                }
                Console.WriteLine("**************************** db call executed *****************************");
                connection.Close();
            }
        }
    }
}
```
When you execute auto instrumentation over it will be changed to something like this
```
using Microsoft.AspNetCore.Mvc;
namespace Samples.AspNetCoreMvc.Controllers
{
    public class HomeController : Controller
    {
        public IActionResult Index()
        {
            StartSpanAndTryToAttachDataFromRequestHeader
            
            return CallDatabase();
            
            StopSpan()
        }
        
        private static void CallDatabase()
        {
            string sql = "select table_catalog, table_schema, table_name from information_schema.tables;";
            using (var connection = new NpgsqlConnection(connString))
            {
                try
                {
                    AttachSpanToExistingTrace()
                    var output = connection.Execute(sql, null, null, null, System.Data.CommandType.Text);
                    StopSpan();
                }
                catch (Exception e)
                {
                    Console.WriteLine(e.ToString());
                    throw;
                }
                Console.WriteLine("**************************** db call executed *****************************");
                connection.Close();
            }
        }
    }
}
```
I hope it answers all doubts from the customer.

Kielek (approved):

lachmatt:
Updated to include most of the @Kielek suggestions. Documentation to be updated with a separate PR.

tracer/src/Datadog.Trace/ClrProfiler/ServerlessInstrumentation/Serverless.cs
```diff
@@ -0,0 +1,142 @@
...
+using Datadog.Trace.Logging;
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace.ClrProfiler.ServerlessInstrumentation
+{
+    internal static class Serverless
+    {
+        private const string DefinitionsId = "68224F20D001430F9400668DD25245BA";
+        private const string ExtensionEnvName = "_SIGNALFX_EXTENSION_PATH";
+        private const string ExtensionFullPath = "/opt/extensions/datadog-agent";
```
pellared: What is this path used for? Is it correct?

tracer/src/Datadog.Trace/Tracer.cs
```diff
@@ -128,19 +130,34 @@ public static Tracer Instance
...
+                lock (GlobalInstanceLock)
+                {
+                    if (_globalInstanceInitialized)
+                    {
+                        return _instance;
+                    }
+
+                    instance = new Tracer(tracerManager: null); // don't replace settings, use existing
+                    _instance = instance;
+                    _globalInstanceInitialized = true;
```
pellared: I think that  `_globalInstanceInitialized` should be `volatile`. Otherwise, there can be a race condition if the code will be reordered in the following way:

```
_globalInstanceInitialized = true;
instance = instance;
```

then the first `if (_globalInstanceInitialized)` can evaluate to `true` before `_instance` is assigned

more: https://www.albahari.com/threading/part4.aspx#_Nonblocking_Synchronization

However, I am also not sure why `_globalInstanceInitialized` field is needed. Probably conditions like `_instance == null` could be used instead and `_globalInstanceInitialized` could be removed. Yet, I may be missing something.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
```diff
@@ -39,8 +39,6 @@ public static void Log(string message, params object[] args)
                         // ignore
                     }
                 }
-
-                Console.Error.WriteLine(message, args);
```
pellared: We should check why it was removed. My guess is that it could be related to https://github.com/signalfx/signalfx-dotnet-tracing/pull/404

pellared (approved):
Few things spotted. None of them looks like a blocker to me.

Kielek (approved):

tracer/src/Datadog.Trace/Tracer.cs
lachmatt: good catch, will be fixed separately

tracer/src/Datadog.Trace/ClrProfiler/ServerlessInstrumentation/Serverless.cs
lachmatt: It's a default path of lambda extension. It is overridden in tests.

PR #401, opened 2022-02-21
Add Windows installer script

RassK:
## What

Adds Windows installer script

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -68,7 +68,34 @@ On Linux, after the installation, you can optionally create the log directory:
 /opt/signalfx/createLogPath.sh
 ```
 
-### Instrument a .NET application on Windows
+## Instrument a .NET application on Windows
+
+Install latest SignalFX .NET agent
+
+```powershell
+# signalfx-dotnet-tracing github repository API
+$api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+
+# determine OS architecture
+$os_bits = (Get-WmiObject Win32_OperatingSystem).OSArchitecture
+$os_arch = (&{If($os_bits -eq "64-bit") {"x64"} Else {"x86"}})
+
+# File pattern to search for
+$pattern = "signalfx-dotnet-tracing-*-$os_arch.msi"
+
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | where { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
+$msi = Join-Path $env:temp $download.name
+
+# Download installer MSI to Temp
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
+
+# Install downloaded MSI
+Start-Process msiexec.exe -Wait -ArgumentList "/I $msi /quiet"
+
+# Cleanup
+Remove-Item $msi                                                
+```
 
 Before running the application, set the following environment variables:
 
@@ -93,7 +120,7 @@ $Env:SIGNALFX_ENV = "production"                                  # Set the envi
 dotnet run                                                        # Run your application                                                     
 ```
 
-### Instrument a .NET application on Linux
+## Instrument a .NET application on Linux
 
 Before running the application, set the following environment variables:
 


docs/README.md
```diff
@@ -68,7 +68,34 @@ On Linux, after the installation, you can optionally create the log directory:
 /opt/signalfx/createLogPath.sh
 ```
 
-### Instrument a .NET application on Windows
+## Instrument a .NET application on Windows
+
+Install latest SignalFX .NET agent
```
pellared, line 73: It is `SignalFx Instrumentationy for .NET`

docs/README.md
```diff
@@ -68,7 +68,34 @@ On Linux, after the installation, you can optionally create the log directory:
...
+
+```powershell
```
pellared, line 75: How about placing somewhere under the `Installation` section? E.g. a subsection like `Automated installation script for Windows`? 

@theletterf WDYT?

docs/README.md
```diff
@@ -68,7 +68,34 @@ On Linux, after the installation, you can optionally create the log directory:
...
+# signalfx-dotnet-tracing github repository API
+$api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+
+# determine OS architecture
+$os_bits = (Get-WmiObject Win32_OperatingSystem).OSArchitecture
```
pellared, line 80: I think the common practice is to use `camelCase` for variable names.
I would try running https://github.com/PowerShell/PSScriptAnalyzer against this script

RassK: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_variables?view=powershell-7.2

Just the docs style...

docs/README.md
theletterf: Agree. If we want everyone to use the install script first (probably preferred), I'd move it there under the heading, with a short introduction, and then document the other manual method below.

docs/README.md
theletterf: Isn't it simpler to just provide a `ps1` script and let people download it? For example:

```
(New-Object System.Net.WebClient).DownloadFile("script-download-url", "$env:TEMP\script.ps1"); 
```

docs/README.md
```diff
@@ -68,7 +68,34 @@ On Linux, after the installation, you can optionally create the log directory:
...
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | where { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
+$msi = Join-Path $env:temp $download.name
+
+# Download installer MSI to Temp
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
+
+# Install downloaded MSI
+Start-Process msiexec.exe -Wait -ArgumentList "/I $msi /quiet"
+
```
theletterf, line 95: @RassK I think we should at least declare the basic env variables, and perhaps take the configurable ones as arguments. WDYT? @pellared

theletterf (changes requested):

docs/README.md
RassK: The MSI is setting up all of the common envs

docs/README.md
RassK: If we have a place where to host it?

RassK: ... plus our security policy for hosting such important things.

docs/README.md
theletterf: Yes, though I'd keep the manual download / command instructions in a subsection right after the automated install one.

docs/README.md
pellared: I prefer to keep it as it is.

Why? 

1. Lower attack surface (we would not rely on GitHub "CDN" as our official GDI docs would simply contain WHOLE code)
2. Personally I often first download such "magic" scripts before executing them. The user could easily see what is going on.
3. We can always improve it later if there would be a demand, however e.g. personally I would like to have such script signed (it is possible to sign PowerShell scripts but we would rather have to keep it in some CDN - not GitHub).

docs/README.md
```diff
@@ -68,7 +68,36 @@ On Linux, after the installation, you can optionally create the log directory:
...
+### Automated installation script for Windows
+
+```powershell
+# signalfx-dotnet-tracing github repository API
+$api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+
+# determine OS architecture
+$os_bits = (Get-CimInstance Win32_OperatingSystem).OSArchitecture
+$os_arch = (&{If($os_bits -eq "64-bit") {"x64"} Else {"x86"}})
```
pellared, line 81: this line is "wow" 😮 but fine for me 😉

docs/README.md
```diff
@@ -68,7 +68,36 @@ On Linux, after the installation, you can optionally create the log directory:
 /opt/signalfx/createLogPath.sh
 ```
 
-### Instrument a .NET application on Windows
+## Installation
```
pellared, line 71: this header can be removed - it is already there above.

pellared: I tried to address it here https://github.com/signalfx/signalfx-dotnet-tracing/pull/401/commits/2c33671dcc7a8debcc3aa9d666469faa9d11ae8b

docs/README.md
pellared: I tried to address it here https://github.com/signalfx/signalfx-dotnet-tracing/pull/401/commits/2c33671dcc7a8debcc3aa9d666469faa9d11ae8b

pellared (approved):

Kielek (approved):

docs/README.md
theletterf: Looks great, @pellared !

theletterf (approved):

docs/README.md
```diff
@@ -49,6 +49,38 @@ to receive telemetry data.
...
+
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
+
+# Download installer MSI to Temp
+$msi = Join-Path $env:temp $download.name
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
+
+# Install downloaded MSI
+Start-Process msiexec.exe -Wait -ArgumentList "/I $msi /quiet"
```
pellared, line 75:
```suggestion
Start-Process msiexec.exe -Wait -ArgumentList "/i $msi /quiet /norestart"
```

and also we miss checking the exit code

as far as I remember we should do nothing if `0` is returned. we should request reboot if `3010` is returned. otherwise we should return the error code and hyperlink to https://docs.microsoft.com/en-us/windows/win32/msi/error-codes to get more info


reference:
- https://docs.microsoft.com/en-us/windows/win32/msi/error-codes

docs/README.md
```diff
@@ -49,6 +49,38 @@ to receive telemetry data.
...
+
+# File pattern to search for
+$pattern = "signalfx-dotnet-tracing-*-$os_arch.msi"
+
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
+
+# Download installer MSI to Temp
+$msi = Join-Path $env:temp $download.name
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
```
theletterf, line 72: I'd wrap all our Invoke-WebRequest calls in try...catch blocks, like this:

```suggestion
try { Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi } catch {
      $_.Exception.Response.StatusCode.Value__}
```

theletterf (changes requested):

docs/README.md
pellared: I think it would make sense only if we create a script file that the user would download and use

theletterf (approved):
I'm approving to unlock this, but as discussed, I would not add this script to the GDI UI and the docs yet until we've a solution for downloading and executing everything using a one-liner, like the OTel Collector distro does.

docs/README.md
RassK: Was specially crafted for early PS versions 😆 ... otherwise they have `(condition) ? true : false` in PS7, but WS2012 has only PS5

docs/README.md
RassK: For public facing scripts I would prefer more inline style to maintain readability + copy / paste.
Can be refactored when we do the CDN hosting and hiding the complexity.

docs/README.md
RassK: Windows should restart anyway due system level env variables that are installed with msi. Could do without using some hacks (like restarting explorer.exe, sign off -> sign in, etc), but I wouldn't advice it.

PR #402, opened 2022-02-22
Document OTEL_PROPAGATORS to SIGNALFX_PROPAGATORS mapping

pellared:
## Why

It is a common error that the user thinks that our `B3` is the single-header implementation. The `W3C` is also unclear.

## What

Document OTEL_PROPAGATORS to SIGNALFX_PROPAGATORS mapping

## Tests

Done by @Kielek when testing propagation between SFx .NET and Splunk OTel Node.js


Kielek (approved):

theletterf (approved):
@pellared Unsure about adding this here now. I'd rather use this information when we launch the OTel distro and publish the migration guide.

Kielek:
@theletterf, it is useful when you want to use propagators between apps intrumented by otel (like node js applications) and .NET.

PR #403, opened 2022-02-23
Create Linux release build on GitLab

pellared:
## Why

Adhere to company standards.

## What

Build the Linux release artifacts in GitLab.

## Tests

Info on Slack.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,15 +38,11 @@ linux-build:
 
 # windows-build:
 #   stage: build
-#   # rules:
-#   #   - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+.*/'
 #   tags:
 #     - windows
 #   script:
 #     - tracer\build.ps1 Clean BuildTracerHome PackageTracerHome
 #   artifacts:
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -1,27 +1,48 @@
+workflow:
+  rules:
+    - if: '$CI_COMMIT_TAG =~ /^v/'
```
RassK, line 4: Is it required to be one the root dir?

pellared: You mean the `.gitlab-ci.yml` file?
It is the default location https://docs.gitlab.com/ee/ci/pipelines/settings.html#specify-a-custom-cicd-configuration-file

I think it is good as our other repositories use this location as well. I think it will be also easier if someone would test things on forks if needed.

RassK (commented):
One more thing, should we keep GH workflow until GL is finalized?
Otherwise if we need to publish something before GL is finalized, there seems to be manual steps to follow (eg missing linux packages in GH).

pellared:
> One more thing, should we keep GH workflow until GL is finalized?

Done.

RassK (approved):

PR #404, opened 2022-02-23
StartupLogger exceptions

dszmigielski:
## Why

StartupLogger logs the dll loading more than once, and we don't wanna bloat the logs, we also want to know why this happens.

## What

We are using AssemblyResolve event for loading our tracer dlls, when they are not in the default context (on failed loading we search for them in the tracer home directory). This event is triggered multiple times for the same dll for some reason (and it seems to be working as intended, as I was able to find some best practices on using the event and it actually mentions that), but we don't have to load the file again. It is enough to return previously loaded assembly and this is what has been done in this PR.

## Tests

CI tests passing

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
@@ -70,9 +70,7 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
             {
                 StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);
                 var loadedAssembly = DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
-                LoadedAssemblies[args.Name] = loadedAssembly;
-
-                return loadedAssembly;
+                return LoadedAssemblies.GetOrAdd(args.Name, loadedAssembly);
             }
 
             return null;
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
@@ -37,6 +37,12 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
 
         private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sender, ResolveEventArgs args)
         {
+            if (LoadedAssemblies.ContainsKey(args.Name))
+            {
+                StartupLogger.Debug("{0} has already been loaded.", args.Name);
+                return LoadedAssemblies[args.Name];
+            }
+
             var assemblyName = new AssemblyName(args.Name);
 
             // On .NET Framework, having a non-US locale can cause mscorlib
@@ -63,7 +69,10 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
             if (File.Exists(path))
             {
                 StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);
-                return DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
+                var loadedAssembly = DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
+                LoadedAssemblies[args.Name] = loadedAssembly;
+
+                return loadedAssembly;
             }
 
             return null;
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs
```diff
@@ -49,7 +55,10 @@ private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sende
                 }
 
                 StartupLogger.Debug("Resolving {0}, loading {1}", args.Name, path);
-                return Assembly.LoadFrom(path);
+                var loadedAssembly = Assembly.LoadFrom(path);
+                LoadedAssemblies[args.Name] = loadedAssembly;
```
pellared, line 59: there can be a race condition that would result in loading the same assembly twice.
I think you would need to use the `GetOrAdd()` method to avoid it

pellared (changes requested):
https://github.com/signalfx/signalfx-dotnet-tracing/pull/404#discussion_r813171658

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs
dszmigielski: Done

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -63,7 +69,8 @@ private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sende
             if (File.Exists(path))
             {
                 StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);
-                return DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
+                var loadedAssembly = DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
+                return LoadedAssemblies.GetOrAdd(args.Name, loadedAssembly);
```
pellared, line 73: Maybe use `GetOrAdd(TKey, Func<TKey,TValue>)` to execute the whole logic. Then you can even remove `LoadedAssemblies.ContainsKey(args.Name)`

The current code suffers from following issues:
1. race condition: You can get `StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);` twice for the same assembly even if was loaded only once
2. performance: It will always execute `File.Exists(path)` if the file does not exist.

pellared:
> This event is triggered multiple times for the same dll for some reason (and it seems to be working as intended, as I was able to find some best practices on using the event and it actually mentions that), but we don't have to load the file again. It is enough to return previously loaded assembly and this is what has been done in this PR.



1. What is the "some reason"? _I have a guess that in parallel multiple threads are trying to load the same assembly..._
2. Is it possible that this code would run in multiple AppDomains? (Answered by @Kielek) yes, but the static fields are unique so it should work.
3. Do we know why we see it triggering more often than upstream? E.g. do we have more concurrent threads?

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
dszmigielski: I changed it to simple Dictionary and added double-checked locking.

dszmigielski:
The stack trace where the event is called is always the same:
trace

I have found a few information on the internet mentioning that this behaviour (event being fired multiple times for the same assembly) is intentional. [Here](https://www.codeproject.com/Articles/310675/AppDomain-AssemblyResolve-Event-Tips) are some best practices for using the `AssemblyResolved` event, including this sentence:

> AssemblyResolve event can fire multiple times for the same assembly.

From what I found the most probable cause is the calling assembly being loaded in the different context that the target assembly, causing the assembly resolver to fire the event. [Here](https://social.msdn.microsoft.com/Forums/en-US/47c997f9-fbde-43f6-be3a-3525459148fd/what-happen-after-the-assembly-returned-from-assemblyresolve-event-handler?forum=asparchitecture) you can read:

> The same assembly can be in the AppDomain multiple times, under the 3 different binding contexts. It a little know fact, but if you have two assemblies, A and B, and A has a dependency to B, when you get a type from A that needs B it will turn to the AssemblyResolver. If A and B were loaded into the AppDomain under different binding contexts, the resolver fails. References between assemblies are resolved within the same context (unless you subscribe to the AssemblyResolver event to help it out).

[Here](https://web.archive.org/web/20190124235454/https://blogs.msdn.microsoft.com/suzcook/2003/05/29/choosing-a-binding-context/) you can find some more information on the different assembly binding contexts.

Kielek:
Probably will be addressed by upstream. Closing for now.

PR #413, opened 2022-03-01
Remove AppSec

RassK:
## What

Removing AppSec as no use for us. Saves build and test time.


pjanotti:
> Saves build and test time.

How much?

Q.: Can we remove from build and packaging instead of removing the sources? Just considering what would be easier for upstream pull.

RassK:
> > Saves build and test time.
> 
> How much?
> 
> Q.: Can we remove from build and packaging instead of removing the sources? Just considering what would be easier for upstream pull.

Not sure about the saved time, as almost all of the logic it is inside `Datadog.Trace`, there is no clear separation and too many rows to individually find the test runtimes. I'm not considering unit tests as a big win but integration tests are definitely something worth.

I think it's not so big modification that would affect upstream pulls.
99% of the logic was inside `Datadog.Trace` AppSec/* dir
Unit tests and integration tests were separated to different library.

pellared (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/packages.config
```diff
@@ -1,4 +1,3 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="libddwaf" version="1.0.17" targetFramework="native" />
```
Kielek, line 3: Can you remove this file?
It is empty, and based what I see references to it are removed.

tracer/src/WindowsInstaller/Product.wxs
```diff
@@ -114,10 +114,6 @@
               Checksum="yes">
           <Class Id="$(var.ProfilerCLSID)" Context="InprocServer32" ThreadingModel="both" Description="$(var.ProductNamePlatformAgnostic)"/>
         </File>
-        <File Id="LibDdwaf"
```
Kielek, line 117: please check upgrade scenario. v0.2.0 to version build with this changes. ddwaf file should be removed by the installer.

Kielek (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/packages.config
RassK: Yep, good catch, seems deleting directly from VS didn't work.

pjanotti (approved):

tracer/src/WindowsInstaller/Product.wxs
RassK: verified on my lab machine. v0.2.2 was installed. built v0.2.3 msi and after upgrade install ddwaf.dll was removed.

PR #421, opened 2022-03-09
Alwayson profiller resharper cleanup

Kielek:
## Why & What

R# static analysis cleanups.
There will be one more follow up PR related to narrowing conversions.

## Tests
CI

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -622,11 +622,11 @@ DWORD WINAPI SamplingThreadMain(_In_ LPVOID param)
     }
 }
 
-void ThreadSampler::StartSampling(ICorProfilerInfo10* info10)
+void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
 {
     Logger::Info("ThreadSampler::StartSampling");
-    profilerInfo = info10;
-    this->info10 = info10;
+    profilerInfo = cor_profiler_info10;
+    this->info10 = cor_profiler_info10;
 #ifdef _WIN32
     HANDLE bgThread = CreateThread(nullptr, 0, &SamplingThreadMain, this, 0, nullptr);
 #else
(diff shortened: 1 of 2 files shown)


pellared (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -135,21 +134,21 @@ ThreadSamplesBuffer ::~ThreadSamplesBuffer()
     buffer = nullptr; // specifically don't delete as this is done by RecordProduced/ConsumeOneThreadSample
 }
 
-#define CHECK_SAMPLES_BUFFER_LENGTH() {  if (buffer->size() >= SAMPLES_BUFFER_MAXIMUM_SIZE) { return; } }
+#define CHECK_SAMPLES_BUFFER_LENGTH() {  if (buffer->size() >= samples_buffer_maximum_size) { return; } }
```
pellared, line 137: how about adding a blank line afterwards?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -604,36 +602,33 @@ void SleepMillis(int millis) {
...
         } else {
             PauseClrAndCaptureSamples(ts, info10, helper);
         }
     }
-
-    return 0;
 }
 
-void ThreadSampler::StartSampling(ICorProfilerInfo10* info10)
+void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
```
RassK, line 625: seems to create naming style conflict with global constants.

lachmatt (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: the whole file has a lot of issues with naming convention.

I am not sure which one should be used. This one is recommended by IDE.

PR #425, opened 2022-03-10
AlwaysOn Profiller - shared generic classes and internal classes

Kielek:
## Why

Extend the AlwaysOn Profiler.

## What

Support for internal classed and shared generic functions.
Cache for class names is removed. Full function names extended about the class name size.

## Tests

CI

## TODO
You can expect more PRs for handling generic parameters.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -9,10 +9,7 @@
   #include <pthread.h>
 #endif
 
-constexpr auto max_func_name_len = 256UL;
-constexpr auto max_class_name_len = 512UL;
 constexpr auto max_string_length = 512UL;
-
 
 constexpr auto max_codes_per_buffer = 10 * 1000;
 
@@ -28,8 +25,7 @@ constexpr auto minimum_sample_period = 1000;
 // FIXME make configurable (hidden)?
 // These numbers were chosen to keep total overhead under 1 MB of RAM in typical cases (name lengths being the biggest
 // variable)
-constexpr auto max_function_name_cache_size = 6000;
-constexpr auto max_class_name_cache_size = 1000;
+constexpr auto max_function_name_cache_size = 7000;
 
 // If you squint you can make out that the original bones of this came from sample code provided by the dotnet project:
 // https://github.com/dotnet/samples/blob/2cf486af936261b04a438ea44779cdc26c613f98/core/profiling/stacksampling/src/sampler.cpp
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -366,43 +304,43 @@ class SamplingHelper
...
         }
 
-        // If the ClassID returned from GetFunctionInfo is 0, then the function
-        // is a shared generic function.
-        if (classId != 0)
-        {
-            LookupClassName(classId, result);
-        }
-        else
+        if (function_info.type.parent_type != nullptr)
```
pjanotti, line 320: For which cases `parent_type` comes into the picture?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -246,13 +242,12 @@ class SamplingHelper
     // These are permanent parts of the helper object
     ICorProfilerInfo10* info10 = nullptr;
     NameCache functionNameCache;
-    NameCache classNameCache;
     // These cycle every sample and/or are owned externally
     ThreadSamplesBuffer* curWriter = nullptr;
     std::vector<unsigned char>* curBuffer = nullptr;
     SamplingStatistics stats;
 
-    SamplingHelper() : functionNameCache(max_function_name_cache_size), classNameCache(max_class_name_cache_size)
```
pjanotti, line 255: 👍

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
pjanotti: Is it for the case like `My.Custom.Test.Namespace.ClassA.InternalClassB.DoubleInternalClassB`? Perhaps it is worth a comment to make that clear.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AlwaysOnProfilerTests.cs
```diff
@@ -105,8 +105,8 @@ private static bool ContainStackTraceForClassHierarchy(LogRecord logRecord)
             return logRecord.Body.StringValue.Contains(
                 "\tat System.Threading.Thread.Sleep(System.TimeSpan)\n" +
                 "\tat My.Custom.Test.Namespace.ClassD`1.GenericMethodDFromGenericClass(!0, !!0)\n" +
-                "\tat SharedGenericFunction.GenericMethodCFromGenericClass(!0)\n" +
-                "\tat TripleInternalClassB.MethodB(System.String)\n" +
+                "\tat My.Custom.Test.Namespace.GenericClassC`1.GenericMethodCFromGenericClass(!0)\n" +
```
pjanotti, line 108: This is definitely looks in line with the symbols that we typically see for .NET 👍

pjanotti (approved):
LGTM

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: Yes, this code is related to internal classes. Comment added.

PR #430, opened 2022-03-14
Lock ASP.NET Core versions

RassK:
## What

* Locks ASP.NET Core versions in dockerfiles.

### DIFF
--- a/tracer/build/_build/docker/alpine.dockerfile
+++ b/tracer/build/_build/docker/alpine.dockerfile
@@ -35,10 +35,10 @@ FROM base as tester
 # Install .NET Core runtimes using install script
 RUN curl -sSL https://dot.net/v1/dotnet-install.sh --output dotnet-install.sh \
     && chmod +x ./dotnet-install.sh \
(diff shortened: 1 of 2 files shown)


pellared (approved):
> [TODO] Adds doc that tracks dependency types to update.

Can be as a separate PR.

tracer/build/_build/docker/alpine.dockerfile
```diff
@@ -35,10 +35,10 @@ FROM base as tester
 # Install .NET Core runtimes using install script
 RUN curl -sSL https://dot.net/v1/dotnet-install.sh --output dotnet-install.sh \
```
pellared, line 36: If I understand [this](https://github.com/signalfx/gdi-specification/blob/33f1b65633c5f83acc1622bbc79f498e513a6c7c/specification/repository.md?plain=1#L48-L49) correctly it would mean that we should vendor (put to the repo) the script instead of downloading it

@owais Am I correct?

owais: Either vendor it or verify with a checksum after download and store the checksum in this file.

RassK: Let's test the checksum option in different PR

PR #431, opened 2022-03-15
Better hresult logs

Kielek:
## Why

Common way to log hresult.

## What

Log hresult as a hex.
Based on https://github.com/signalfx/signalfx-dotnet-tracing/blob/7862472b59eb1419bd8ae43df903c950f11df3d4/tracer/src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp#L46-L47

Opportunistic fix for interface id.

## Tests

CI


pjanotti (approved):
The preference would be to avoid the boiler plate code somehow but if that is not possible or advisable for some reason LGTM.

PR #438, opened 2022-03-21
Extract type parameter names

lachmatt:
## Why

Include type parameter names in stacktraces.

## What

Cloned and modified `GetSigTypeTokName` from `clr_helpers.cpp` to extract type parameter names.

## Tests

Existing tests for AlwaysOnProfiling.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -27,6 +27,10 @@ constexpr auto minimum_sample_period = 1000;
 // variable)
 constexpr auto max_function_name_cache_size = 7000;
 
+constexpr auto param_name_max_len = 260;
+constexpr auto generic_params_max_len = 20;
+constexpr auto unknown = WStr("Unknown");
+
 // If you squint you can make out that the original bones of this came from sample code provided by the dotnet project:
 // https://github.com/dotnet/samples/blob/2cf486af936261b04a438ea44779cdc26c613f98/core/profiling/stacksampling/src/sampler.cpp
 // That stack sampling project is worth reading for a simpler (though higher overhead) take on thread sampling.
@@ -279,6 +283,7 @@ private:
         constexpr auto unknown_list_of_arguments = WStr("(unknown)");
         constexpr auto unknown_function_name = WStr("Unknown(unknown)");
         constexpr auto name_separator = WStr(".");
+
         if (funcID == 0)
         {
             constexpr auto unknown_native_function_name = WStr("Unknown_Native_Function(unknown)");
@@ -288,10 +293,10 @@ private:
 
         ClassID classId = 0;
         ModuleID moduleId = 0;
-        mdToken token = 0;
+        mdToken functionToken = 0;
 
         // theoretically there is a possibility to use GetFunctionInfo method, but it does not support generic methods
-        HRESULT hr = info10->GetFunctionInfo2(funcID, frameInfo, &classId, &moduleId, &token, 0, nullptr, nullptr);
+        HRESULT hr = info10->GetFunctionInfo2(funcID, frameInfo, &classId, &moduleId, &functionToken, 0, nullptr, nullptr);
         if (FAILED(hr))
         {
             Logger::Debug("GetFunctionInfo2 failed. HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
@@ -308,7 +313,7 @@ private:
             return;
         }
 
-        const auto function_info = GetFunctionInfo(pIMDImport, token);
+        const auto function_info = GetFunctionInfo(pIMDImport, functionToken);
 
         if (!function_info.IsValid())
         {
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -337,17 +343,36 @@ class SamplingHelper
...
+        HCORENUM functionGenParamsIter = nullptr;
+        HCORENUM classGenParamsIter = nullptr;
+        mdGenericParam functionGenericParams[generic_params_max_len]{};
+        mdGenericParam classGenericParams[generic_params_max_len]{};
+        ULONG functionGenParamsCount = 0;
+        ULONG classGenParamsCount = 0;
+
+        mdTypeDef classToken = function_info.type.id;
+
+        pIMDImport->EnumGenericParams(&classGenParamsIter, classToken, classGenericParams, generic_params_max_len,
```
Kielek, line 356: consider `hr = pIMDImport`
and then  if (FAILED(hr))

both for method and the class

Consider the same behaviour as for failing functionMethodSignature.TryParse

Kielek (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
lachmatt: fixed

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -356,12 +395,167 @@ class SamplingHelper
...
+        ULONG num = 0;
+        pbCur += CorSigUncompressData(pbCur, &num);
+        if (num >= generic_params_max_len)
+        {
+            return unknown;
+        }
+        WCHAR param_type_name[param_name_max_len]{};
+        ULONG pch_name = 0;
+        const auto hr = pImport->GetGenericParamProps(genericParameters[num], nullptr, nullptr, nullptr, nullptr,
+                                                      param_type_name, param_name_max_len - 1, &pch_name);
```
pjanotti, line 418: not from this PR itself but we need a consistent way to treat calls that break over multiple lines. I don't think we are being consistent about them.

pjanotti: `param_name_max_len - 1`: doesn't the API already handles the space for the terminating `\0`?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -356,12 +395,167 @@ class SamplingHelper
...
+            {
+                pbCur++;
+                tokenName = GetSigTypeTokName(pbCur, pImport, classParams, methodParams) + WStr("[]");
+                break;
+            }
+            case ELEMENT_TYPE_GENERICINST:
+            {
+                pbCur++;
+                tokenName = GetSigTypeTokName(pbCur, pImport, classParams, methodParams);
+                tokenName += WStr("[");
```
pjanotti, line 524: Isn't more typical to use angle brackets here?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -337,17 +343,50 @@ class SamplingHelper
...
+        mdGenericParam functionGenericParams[generic_params_max_len]{};
+        mdGenericParam classGenericParams[generic_params_max_len]{};
+        ULONG functionGenParamsCount = 0;
+        ULONG classGenParamsCount = 0;
+
+        mdTypeDef classToken = function_info.type.id;
+
+        hr = pIMDImport->EnumGenericParams(&classGenParamsIter, classToken, classGenericParams, generic_params_max_len,
+                                     &classGenParamsCount);
+        pIMDImport->CloseEnum(classGenParamsIter);
```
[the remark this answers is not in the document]
lachmatt: yes

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -337,17 +343,50 @@ class SamplingHelper
...
+            Logger::Debug("Method generic parameters enumeration failed. HRESULT=0x", std::setfill('0'), std::setw(8),
+                          std::hex, hr);
+            result.append(unknown_list_of_arguments);
+            return;
+        }
+        
         // try to list arguments type
-        FunctionMethodSignature function_method_signature = function_info.method_signature;
-        hr = function_method_signature.TryParse();
+        FunctionMethodSignature functionMethodSignature = function_info.method_signature;
```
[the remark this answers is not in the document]
lachmatt: Our naming is inconsistent. I tried to make it more consistent, but decided to postpone it.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
lachmatt: This was copied from `clr_helpers.cpp` as is, but I can adjust it.

Kielek: After internal discussions, we will keep square brackets. It it aligned with stacktraces.

PR #439, opened 2022-03-22
Always on better format fot generic methods

Kielek:
## Why & What

 - fixes to comment form #438
 - display method generic parameters name
 - Extended test case to cover scenario where we have more than 20 generic parameter names

## Tests

CI/New test created.

## Notes

After internal conversation we have decided to do not list all generic parameters from the classes and keep only the number of it. It is same as we have in stacktraces.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -363,11 +363,11 @@ private:
             result.append(unknown_list_of_arguments);
             return;
         }
+        
         hr = pIMDImport->EnumGenericParams(&functionGenParamsIter, functionToken, functionGenericParams,
                                      generic_params_max_len,
                                       &functionGenParamsCount);
         pIMDImport->CloseEnum(functionGenParamsIter);
-
         if (FAILED(hr))
         {
             Logger::Debug("Method generic parameters enumeration failed. HRESULT=0x", std::setfill('0'), std::setw(8),
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -415,7 +415,7 @@ private:
         WCHAR param_type_name[param_name_max_len]{};
         ULONG pch_name = 0;
         const auto hr = pImport->GetGenericParamProps(genericParameters[num], nullptr, nullptr, nullptr, nullptr,
-                                                      param_type_name, param_name_max_len - 1, &pch_name);
+                                                      param_type_name, param_name_max_len, &pch_name);
         if (FAILED(hr))
         {
             Logger::Debug("GetGenericParamProps failed. HRESULT=0x", std::setfill('0'), std::setw(8),
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -375,6 +375,34 @@ private:
             result.append(unknown_list_of_arguments);
             return;
         }
+
+        if (functionGenParamsCount > 0)
+        {
+            result.append(WStr("["));
+            for (ULONG i = 0; i < functionGenParamsCount; ++i)
+            {
+                if (i != 0)
+                {
+                    result.append(WStr(", "));
+                }
+
+                WCHAR param_type_name[param_name_max_len]{};
+                ULONG pch_name = 0;
+                const auto hr =
+                    pIMDImport->GetGenericParamProps(functionGenericParams[i], nullptr, nullptr, nullptr, nullptr,
+                                                     param_type_name, param_name_max_len, &pch_name);
+                if (FAILED(hr))
+                {
+                    Logger::Debug("GetGenericParamProps failed. HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
+                    result.append(unknown);
+                }
+                else
+                {
+                    result.append(param_type_name);
+                }
+            }
+            result.append(WStr("]"));
+        }
         
         // try to list arguments type
         FunctionMethodSignature functionMethodSignature = function_info.method_signature;
--- a/tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
+++ b/tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
@@ -78,7 +78,7 @@ namespace My.Custom.Test.Namespace
             CustomStruct[] structArray,
             List<T> genericList)
         {
-            void Action(string s) => InternalClassB<string, int>.DoubleInternalClassB.TripleInternalClassB<int>.MethodB(s, new int[] { 3 }, TimeSpan.Zero, 0, new List<string>{"a"});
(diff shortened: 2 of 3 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -363,18 +363,46 @@ class SamplingHelper
...
         {
             Logger::Debug("Method generic parameters enumeration failed. HRESULT=0x", std::setfill('0'), std::setw(8),
                           std::hex, hr);
             result.append(unknown_list_of_arguments);
             return;
         }
+
+        if (functionGenParamsCount > 0)
+        {
+            result.append(WStr("["));
```
pjanotti, line 381: nit: please define these as constants so we can easily change if needed.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -363,18 +363,46 @@ class SamplingHelper
...
+            for (ULONG i = 0; i < functionGenParamsCount; ++i)
+            {
+                if (i != 0)
+                {
+                    result.append(WStr(", "));
```
pjanotti, line 386: Isn't this appending an extra ", " after the last parameter type?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -363,18 +363,46 @@ class SamplingHelper
...
+                }
+
+                WCHAR param_type_name[param_name_max_len]{};
+                ULONG pch_name = 0;
+                const auto hr =
+                    pIMDImport->GetGenericParamProps(functionGenericParams[i], nullptr, nullptr, nullptr, nullptr,
+                                                     param_type_name, param_name_max_len, &pch_name);
+                if (FAILED(hr))
+                {
+                    Logger::Debug("GetGenericParamProps failed. HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
```
pjanotti, line 396: Something to investigate perhaps: do we really want to keep trying after first failure? Not to be resolved now.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
pjanotti: also const too so we can see the characters that we are using to build the symbol name in a single location

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -363,18 +363,46 @@ class SamplingHelper
...
+                    result.append(unknown);
+                }
+                else
+                {
+                    result.append(param_type_name);
+                }
+            }
+            result.append(WStr("]"));
```
pjanotti, line 404: Same as above: const.

tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
```diff
@@ -101,11 +101,28 @@ internal static class GenericClassC<T>
     {
         public static void GenericMethodCFromGenericClass(T arg)
         {
-            ClassD<TimeSpan>.GenericMethodDFromGenericClass(TimeSpan.MaxValue, arg, 1);
+            GenericMethodCFromGenericClass(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21);
+        }
+
+        public static void GenericMethodCFromGenericClass<T01, T02, T03, T04, T05, T06, T07, T08, T09, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21>(T01 p01, T02 p02, T03 p03, T04 p04, T05 p05, T06 p06, T07 p07, T08 p08, T09 p09, T10 p10, T11 p11, T12 p12, T13 p13, T14 p14, T15 p15, T16 p16, T17 p17, T18 p18, T19 p19, T20 p20, T21 p21)
+        {
+            // Always on profiler supports only fetching for 20 generic arguments. This method covers scenario where there is more parameters.
```
pjanotti, line 109:
```suggestion
            // Always on profiler supports fetching at most 20 generic arguments. This method covers scenario where there are more than 20 parameters.
```

tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
```diff
@@ -101,11 +101,28 @@ internal static class GenericClassC<T>
...
+            ClassD<int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int>.MethodD(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21);
+        }
+
+    }
+
+    internal static class ClassD<T01, T02, T03, T04, T05, T06, T07, T08, T09, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21>
+    {
+        // Always on profiler supports only fetching for 20 generic arguments. This class covers scenario where there is more parameters.
```
pjanotti, line 117:
```suggestion
            // Always on profiler supports fetching at most 20 generic arguments. This method covers scenario where there are more than 20 parameters.
```

pjanotti (commented):
Could you please checking the extra ", "?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: The `, `  is not added for the 0-ish parameter.
So if we have one parameter it will produce just `<T>` If 2 -> `<T1, T2>`, etc.

tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
Kielek: a788146bc1ea8c81f089b363f1075a48a8714afd

tracer/test/test-applications/integrations/Samples.AlwaysOnProfiler/Program.cs
Kielek: a788146bc1ea8c81f089b363f1075a48a8714afd

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: a788146bc1ea8c81f089b363f1075a48a8714afd

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: a788146bc1ea8c81f089b363f1075a48a8714afd

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: a788146bc1ea8c81f089b363f1075a48a8714afd

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: Well, I think that the case where is could be frequent is handled by 277bee7dc316629c579aa3de062f55ad7ba88e2f
I do not see any other obvious cases which can fails frequently.

lachmatt (approved):

PR #451, opened 2022-04-04
AlwaysOn Profiler - cache by function token

Kielek:
## Why

FunctionId is volatile and can be changed.

## What

Cache for function name based on Function Token and Module Id.

## Tests

CI
Local tests  to verify if the cache is working correctly. For our Sample app each function name is calculated only once. Second sampling is taking data from the cache.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -284,33 +284,50 @@ public:
     }
 
 private:
-    void GetFunctionName(FunctionID funcID, const COR_PRF_FRAME_INFO frameInfo, shared::WSTRING& result)
+    [[nodiscard]] FunctionIdentifier GetFunctionIdentifier(const FunctionID func_id, const COR_PRF_FRAME_INFO frame_info) const
+    {
+        if (func_id == 0)
+        {
+            constexpr auto zero_valid_function_identifier = FunctionIdentifier{0, 0, true};
+            return zero_valid_function_identifier;
+        }
+
+        ModuleID moduleId = 0;
+        mdToken functionToken = 0;
+        // theoretically there is a possibility to use GetFunctionInfo method, but it does not support generic methods
+        HRESULT hr = info10->GetFunctionInfo2(func_id, frame_info, nullptr, &moduleId, &functionToken, 0, nullptr, nullptr);
+        if (FAILED(hr))
+        {
+            Logger::Debug("GetFunctionInfo2 failed. HRESULT=0x", std::setfill('0'), std::setw(8), std::hex, hr);
+            constexpr auto zero_invalid_function_identifier = FunctionIdentifier{0, 0, false};
+            return zero_invalid_function_identifier;
+        }
+
+        return FunctionIdentifier{functionToken, moduleId, true};
+    }
+
+    void GetFunctionName(FunctionIdentifier functionIdentifier, shared::WSTRING& result)
     {
         constexpr auto unknown_list_of_arguments = WStr("(unknown)");
         constexpr auto unknown_function_name = WStr("Unknown(unknown)");
         constexpr auto name_separator = WStr(".");
 
-        if (funcID == 0)
+        if (!functionIdentifier.is_valid)
+        {
+            result.append(unknown_function_name);
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -284,41 +284,58 @@ class SamplingHelper
     }
 
 private:
-    void GetFunctionName(FunctionID funcID, const COR_PRF_FRAME_INFO frameInfo, shared::WSTRING& result)
+    [[nodiscard]] FunctionIdentifier GetFunctionIdentifier(const FunctionID func_id, const COR_PRF_FRAME_INFO frame_info) const
```
pjanotti, line 287: I learned something :) I wasn't aware of this `[[nodiscard]]`, nice.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -284,41 +284,58 @@ class SamplingHelper
...
+
+        if (!functionIdentifier.is_valid)
+        {
             result.append(unknown_function_name);
             return;
         }
 
+        if (functionIdentifier.function_token == 0)
+        {
+            constexpr auto unknown_native_function_name = WStr("Unknown_Native_Function(unknown)");
```
pjanotti, line 323: nit: I think here it could be dynamic code, until we are certain it is incorrect to say native.
```suggestion
            constexpr auto unknown_native_function_name = WStr("Unknown_Metadata_Token(unknown)");
```

pjanotti (approved):
LGTM, just some nits. We should add to the backlog stories to for dynamic code, or more in general managed code without mdToken. As discussed it seems a low priority issue let's record it for possible future work.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: For now, I will keep as it is. I have checked 
```
    internal static class ClassENonStandardCharactersĄĘÓŁŻŹĆąęółżźśćĜЖᏳⳄʤǋₓڿଟഐቐ〣‿؁੮ᾭ_<TClass>
    {
        public static void GenericMethodDFromGenericClass<TMethod, TMethod2>(TClass classArg, TMethod methodArg, TMethod2 additionalArg)
        {
            dynamic test = new Test();

            test();

        }
    }

    internal class Test : DynamicObject
    {

        public override bool TryInvoke(InvokeBinder binder, object[] args, out object result)
        {
            ClassVb.MethodVb("testParam");
            result = null;
            return true;
        }
    }
```

and it produces `dev-otel-collector-1  |         at My.Custom.Test.Namespace.Test.TryInvoke(System.Dynamic.InvokeBinder, System.Object[], System.Object&)`

Also `dev-otel-collector-1  |         at Unknown_Native_Function(unknown)` is always produces as the root of the stack trace. Please let me know what is scenario you wolud like to cover. I will try to do it or create separate story.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -284,41 +284,58 @@ class SamplingHelper
...
+        constexpr auto unknown_function_name = WStr("Unknown(unknown)");
+        constexpr auto name_separator = WStr(".");
+
+        if (!functionIdentifier.is_valid)
+        {
             result.append(unknown_function_name);
             return;
         }
 
+        if (functionIdentifier.function_token == 0)
```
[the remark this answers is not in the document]
Kielek: I would like to cover it together with the next one comment.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
Kielek: If you are using Visual Studio it is great to install ReSharper for C++.
I am not familiar with all new C++ features so it is suggesting it.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
pjanotti: This is fine, we will dig into later.

PR #456, opened 2022-04-06
Alwaysonprofiler do not run on old env

Kielek:
## Why

To have a stable execution on old frameworks.

## What

Only log errors when AlwaysOn Profiler is enabled on old frameworkds.

## Tests

CI
Manualy executed `Samples.AlwaysOnProfiler.Performance` for all target frameworks.


tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -247,13 +249,18 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
+    if (IsThreadSamplingEnabled())
+    {
+        if (metThreadSamplingRequirements)
+        {
             this->threadSampler = new ThreadSampler();
             this->threadSampler->StartSampling(info10);
         }
+        else
+        {
+            Logger::Error("Could not enable thread sampling: CLR version not compatible");
```
pjanotti, line 262:
```suggestion
            Logger::Error("Could not enable thread sampling: CLR version not supported");
```

pjanotti (approved):
In the future if more requirements to get stack sampling are added we may want to combine both under a single method but for current requirements it LGTM.

PR #459, opened 2022-04-08
AlwaysOn Profiler - Native code naming convention 

Kielek:
## Why & What

Native code naming convention based on Google recommendations

https://google.github.io/styleguide/cppguide.html#Naming
Version delivered by ReSharper C++

## Tests

CI

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -10,94 +10,94 @@
   #include <codecvt>
 #endif
 
-constexpr auto max_string_length = 512UL;
+constexpr auto kMaxStringLength = 512UL;
 
-constexpr auto max_codes_per_buffer = 10 * 1000;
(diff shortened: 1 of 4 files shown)


pellared:
Proposals (can be as different PR):

- can you describe how to configure it in Rider/Resharper https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/internal/developing.md ?
- can we make Nuke target for it and run it also on CI? we could use [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html) to achieve it

We can then add it to https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation

pellared (approved):

Kielek:
The easiest way is to commit generated *.sln.DotSettings then it will be shared by the repository by all developers.
I decided to not commit because in on our case it is one time job.
I have cleaned newly created code but other parts should be untouched - as we are making upstream syncs.

If you want to set up this Resharper C++ configuration you should
Extension -> ReSharper -> Manage Options - Solution Team-shared (double click) -> Code Editing => C++ -> Naming Style -> Reset To Scheme -> Google
image

PR #462, opened 2022-04-19
Config switch to sync export

johnbley:
This will be needed for aws_lambda/serverless/azure_function support.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
@@ -104,6 +104,6 @@ namespace Datadog.Trace.Configuration
         /// Configuration key to do synchronous export of traces.
         /// Default is <c>false</c>
         /// </summary>
-        public const string TraceSynchExport = "SIGNALFX_TRACE_SYNCH_EXPORT";
+        public const string TraceSyncExport = "SIGNALFX_TRACE_SYNC_EXPORT";
     }
 }
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
@@ -99,5 +99,11 @@ namespace Datadog.Trace.Configuration
         /// </summary>
         /// <seealso cref="ExporterSettings.PartialFlushMinSpans"/>
         public const string PartialFlushMinSpans = "SIGNALFX_TRACE_PARTIAL_FLUSH_MIN_SPANS";
+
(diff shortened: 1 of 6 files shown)


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.Exporter.cs
```diff
@@ -99,5 +99,11 @@ internal static partial class ConfigurationKeys
         /// </summary>
         /// <seealso cref="ExporterSettings.PartialFlushMinSpans"/>
         public const string PartialFlushMinSpans = "SIGNALFX_TRACE_PARTIAL_FLUSH_MIN_SPANS";
+
+        /// <summary>
+        /// Configuration key to do synchronous export of traces.
+        /// Default is <c>false</c>
+        /// </summary>
+        public const string TraceSynchExport = "SIGNALFX_TRACE_SYNCH_EXPORT";
```
RassK, line 107: `SIGNALFX_TRACE_SYNCH_EXPORT` any particular reason why not sync instead of synch?

pjanotti (commented):
Other than the "h" mentioned by @RassK it looks good.

Q.: Are we going to deploy auto-instrumentation to serverless?

johnbley:
> Other than the "h" mentioned by @RassK it looks good.
I'll fix that today.

> Q.: Are we going to deploy auto-instrumentation to serverless?
The current decision is no based on instability I saw.  We can revisit after we get some basic manual wrappers working.

Kielek (approved):

tracer/src/Datadog.Trace/TraceContext.cs
```diff
@@ -83,7 +83,8 @@ public void AddSpan(Span span)
 
         public void CloseSpan(Span span)
         {
-            bool ShouldTriggerPartialFlush() => Tracer.Settings.ExporterSettings.PartialFlushEnabled && _spans.Count >= Tracer.Settings.ExporterSettings.PartialFlushMinSpans;
+            bool synch = Tracer.Settings.ExporterSettings.SyncExport;
```
RassK, line 86:
```suggestion
            bool isSync = Tracer.Settings.ExporterSettings.SyncExport;
```

tracer/src/Datadog.Trace/TraceContext.cs
```diff
@@ -140,6 +141,11 @@ public void CloseSpan(Span span)
             if (spansToWrite.Count > 0)
             {
                 Tracer.Write(spansToWrite);
+                if (synch)
```
RassK, line 144:
```suggestion
                if (isSync )
```

RassK (approved):

lachmatt (approved):

PR #465, opened 2022-04-20
Upstream pull followup - Reuse upstream propagation code

dszmigielski:
## Why

We don't need our propagation code anymore as it is now available in the upstream repo.

## What

Reuse upstream propagation code which was recently refactored here:  [SpanContextPropagator refactor by tonyredondo · Pull Request #2536 · DataDog/dd-trace-dotnet](https://github.com/DataDog/dd-trace-dotnet/pull/2536)


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -376,6 +368,16 @@ internal static partial class ConfigurationKeys
         /// <seealso cref="TracerSettings.ExpandRouteTemplatesEnabled"/>
         public const string ExpandRouteTemplatesEnabled = "SIGNALFX_TRACE_EXPAND_ROUTE_TEMPLATES_ENABLED";
 
+        /// <summary>
+        /// Configuration key for setting the propagation style injection.
+        /// </summary>
+        public const string PropagationStyleInject = "SINGALFX_PROPAGATION_STYLE_INJECT";
```
RassK, line 374: this is a breaking change. not sure if we should actually do it.
Might be better to leave the original `SIGNALFX_PROPAGATORS`, document these 2 new vars in internal config and assign value from `SIGNALFX_PROPAGATORS`

dszmigielski: Good point, I will preserve our variable name and populate new properties from its value.

pjanotti (changes requested):
I didn't look at it yet, but, let's block merges until pull from upstream is done.

Kielek:
Do you see any option to remove everything under `Datadog.Trace.Propagation` namespace? It is not available on the upstream side.

dszmigielski:
> Do you see any option to remove everything under `Datadog.Trace.Propagation` namespace? It is not available on the upstream side.

What is left there are header names and some extensions that we are still using. I would rather leave them be tbh.

RassK (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -188,6 +186,10 @@ public TracerSettings(IConfigurationSource source)
             DelayWcfInstrumentationEnabled = source?.GetBool(ConfigurationKeys.FeatureFlags.DelayWcfInstrumentationEnabled)
                                             ?? false;
 
+            PropagationStyleInject = TrimSplitString(source?.GetString(ConfigurationKeys.Propagators) ?? nameof(Propagators.ContextPropagators.Names.B3), ',').ToArray();
```
Kielek, line 189: You have the same logic twice. Consider to call it only once.

tracer/test/snapshots/GrpcLegacy.SubmitTraces_httpclient=Disabled.verified.txt
```diff
@@ -2,64 +2,77 @@
   {
     TraceId: Id_1,
     SpanId: Id_2,
+    Name: SendBothStreamingRequest,
```
Kielek, line 5: I am not able to verify this file. Please check if the changes are expected.

Kielek (commented):
One more thing, did you have chance to verify if long TraceId is correctly propagated? As I remember upstream supports only short TraceIds.

PR #467, opened 2022-04-22
Upload windows zip package

RassK:
## What

Windows zip package is generated during the release cycle but never uploaded as an artifact.


pellared (approved):

pjanotti (dismissed):
It actually LGTM, but, let's block merges until pull from upstream is done.

Kielek (approved):

PR #475, opened 2022-04-26
Nanoserver docs

RassK:
## What

Adds instructions how to use nanoserver

## QA
Should we add separate instructions for nanoserver outside of Docker environment? 

_Creating as a WIP as v0.2.3 is not published yet. That release adds zip artifacts._

### DIFF
--- a/docs/internal/windows-nanoserver.md
+++ b/docs/internal/windows-nanoserver.md
@@ -1,0 +1,32 @@
+# Windows nanoserver
+
+Setting up SignalFx Instrumentation for .NET to run in Windows Nanoserver Docker image.
+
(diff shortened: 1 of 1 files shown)


pjanotti (approved):
Nice, simple and straight to the point.

pjanotti:
> Should we add separate instructions for nanoserver outside of Docker environment?

I don't think so - my guess will be that 99.9% of Nano server usage is as a Docker container. One can always get the image and put it somewhere else but for anyone advanced enough to do that the Dockerfile should be more than enough.

Kielek (approved):

lachmatt (approved):

PR #482, opened 2022-05-02
Fix OpenTracing integration bug

dszmigielski:
## Why

If user application tries iterating over all types of all loaded assemblies and does not use `OpenTracing `integration (`OpenTracing` library is not shipped with the application), then the application crashes. This is because we load `SignalFx.Tracing.OpenTracing` library to the app domain (that references `OpenTracing`) when trying to register `OpenTracing` tracer. We catch the exception when doing that, but the library has already been loaded in the app domain. Because of that when iterating over all assemblies in the domain (`AppDomain.CurrentDomain.GetAssemblies()`) exception is thrown in the client app.

## What

Fix the bug crashing the client applications that iterate over all assemblies in the application context.

## Tests

CI passes


tracer/src/Datadog.Trace/Tracer.cs
```diff
@@ -407,16 +407,21 @@ private static void RegisterGlobalTracer(Tracer instance)
...
-                methodInfo?.Invoke(obj: null, args);
+                Assembly.Load(new AssemblyName("OpenTracing, Version=0.12.1.0, Culture=neutral, PublicKeyToken=61503406977abdaf"));
             }
             catch (Exception ex)
             {
                 Log.Information(ex, "OpenTracing integration was not loaded.");
+                return;
             }
+
+            var signalFxOpenTracingAssembly = Assembly.Load(new AssemblyName("SignalFx.Tracing.OpenTracing, Version=0.2.3.0, Culture=neutral, PublicKeyToken=e43a27c2023d388a"));
```
Kielek, line 418: Are you sure, that there is no possibility to throw exception outside try-catch block?
If so, I am fine with changes.

Kielek (approved):

tracer/src/Datadog.Trace/Tracer.cs
dszmigielski: There shouldn't be if we are installed properly. I was actually tempted to leave the second try-catch block but I don't think it is necessary.

PR #499, opened 2022-05-20
[AlwaysOn Profiler] Optimize GetFunctionInfo

Kielek:
## Why

To have simpler version of GetFunctionInfo

## What

Create specific implementation of GetFunctionInfo to reduce unnecessery calls.
Move AlwaysOn Profiler features to separate namespace

## Tests

CI


tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler_clr_helpers.cpp
```diff
@@ -0,0 +1,354 @@
...
+    // parent_token could be: TypeDef, TypeRef, TypeSpec, ModuleRef, MethodDef
+    const auto type_info = GetTypeInfo(metadata_import, parent_token);
+
+    if (is_generic)
+    {
+        return {method_spec_token,
+                shared::WSTRING(function_name),
+                type_info,
+                FunctionMethodSignature(raw_signature, raw_signature_len)};
+    }
```
pellared, line 71: this can be probably removed together with `method_spec_token` and `is_generic` variables

dszmigielski (approved):
LGTM

PR #501, opened 2022-05-24
[AlwaysOn Profiler] Two layers cache for function names

Kielek:
## Why

To have better performance for fetching function name.

## What

Add second layer for the cache.
Size of the cache is arbitrary.

## Tests

CI.
Manual tests execution for `Samples.AlwaysOnProfiler.Performance` for thread sampling time (in micros).

| 2 layers | 1 layer |
| -- | -- |
| 4264 | 3282 |
| 3078 | 2560 |
| 3129 | 2807 |
| 2559 | 4664 |
| 2564 | 5568 |
| 2546 | 3726 |
| 2296 | 2755 |
| 2210 | 2465 |
| 3081 | 2888 |
| 4315 | 4543 |
| 3626 | 4845 |
| 2553 | 2973 |
| 2381 | 2642 |
| 2189 | 3870 |
| 2601 | 2567 |
| 2462 | 2540 |
| 2569 | 2746 |
| 2351 | 2624 |
| 3123 | 2852 |
| 3987 | 4057 |

2 layers total = 57884 
1 layer total = 66974 

(2 layers total - 1 layer total) / 1 layer total *100% ≈ **13.5%**


pjanotti (approved):
LGTM.

A suggestion for the future would be to add some debug log (or even metrics) when maximum sizes are reached so we can potentially tune those. That said this would be extra fine tuning on top the current one.

pellared (approved):

Kielek:
> A suggestion for the future would be to add some debug log (or even metrics) when maximum sizes are reached so we can potentially tune those. That said this would be extra fine tuning on top the current one.

If the debug log is enable, we put information about new functions not available in the 2nd layer of the cache.
For now, it seems to be enough.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -429,6 +429,10 @@ class SamplingHelper
 public:
     shared::WSTRING* Lookup(FunctionID fid, COR_PRF_FRAME_INFO frame)
     {
+        // this method is using two layers cache
```
pellared, line 432:
```suggestion
        // This method is using two layers of caching
```

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -429,6 +429,10 @@ class SamplingHelper
...
+        // 1st layer depends on FunctionID which is volatile (and valid only within one thread suspension)
+        // 2nd layers depends on mdToken for function (which is stable) and ModuleId which could be volatile,
```
pellared, line 434:
```suggestion
        // 2nd layer depends on mdToken for function (which is stable) and ModuleId which could be volatile,
```

PR #502, opened 2022-05-25
Document how to verify file integrity

pellared:
## Why

Add instructions for users if they want to check the signatures.

## What

Document how to verify file integrity

## Tests

TBH I tested only the `checksums.txt` and  `checksums.txt.asc`. I am 80% sure the DEB and RPM package verification is correct. At the same time, I estimate for 80% that nobody will use these docs. I am lazy and I don't think it is worth investing more time on testing it at the moment.

### DIFF
--- a/docs/internal/verify-file-integrity.md
+++ b/docs/internal/verify-file-integrity.md
@@ -1,0 +1,43 @@
+# Verify file integrity
+
+## Checksums
+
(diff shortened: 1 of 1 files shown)


Kielek (approved):
You have my approval, but please execute commands at least once.

PR #525, opened 2022-06-14
Upstream sync 5d8188ad34a60d6195ed8c6e0d6e3a9e68352d3c

Kielek:
## Why

Leftotover from previous upstream sync

## What

5d8188ad34a60d6195ed8c6e0d6e3a9e68352d3c

## Tests

CI

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
@@ -2,8 +2,10 @@
 
 using System;
 using System.Collections.Specialized;
+using System.Linq;
 using Datadog.Trace.Agent;
 using Datadog.Trace.Configuration;
+using Datadog.Trace.SignalFx.Metrics;
 using FluentAssertions;
 using Xunit;
 using MetricsTransportType = Datadog.Trace.Vendors.StatsdClient.Transport.TransportType;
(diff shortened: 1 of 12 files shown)


tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -191,41 +174,217 @@ public void Traces_SocketFilesExist_ExplicitWindowsPipeConfig_UsesWindowsNamedPi
...
+        var param = 9333;
+        var settings = new ExporterSettings { DogStatsdPort = param };
+        var settingsFromSource = Setup($"SIGNALFX_DOGSTATSD_PORT-{param}", "SIGNALFX_METRICS_EXPORTER-StatsD");
+
+        settings.DogStatsdPort.Should().Be(param);
+        settingsFromSource.DogStatsdPort.Should().Be(param);
+
+        CheckDefaultValues(settings, "DogStatsdPort", "MetricsTransport", "MetricsExporter", "MetricsEndpointUrl");
+        CheckDefaultValues(settingsFromSource, "DogStatsdPort", "MetricsTransport", "MetricsExporter", "MetricsEndpointUrl");
+    }
```
pellared, line 265: 1. how about using `nameof` instead of magic string values?
2. how about adding `using new AssertionScope()`?

these comments applies to all new tests

dszmigielski (approved):

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
Kielek: it it mostly upstream sync code. I would like to avoid modifications

PR #526, opened 2022-06-14
[AlwaysOnProfiling][overhead-tests] Dashboard code

lachmatt:
## Why

Visualize overhead test results.

## What

- Copy and adapt [dashboard](https://signalfx.github.io/splunk-otel-java-overhead-test/) code from https://github.com/signalfx/splunk-otel-java-overhead-test
- Include high frequency profiling configuration in test for comparison

## Tests

Tested locally.

### DIFF
--- a/overhead-test/src/Configs/AgentConfig.cs
+++ b/overhead-test/src/Configs/AgentConfig.cs
@@ -1,6 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
 
 namespace SignalFx.OverheadTest.Configs;
 
@@ -10,13 +8,13 @@ internal class AgentConfig
         "none",
         "no agent at all",
         BaseImageName,
-        Enumerable.Empty<KeyValuePair<string, string>>());
+        Array.Empty<EnvVar>() );
 
     public static readonly AgentConfig Instrumented = new(
         "signalfx-dotnet",
         "signalfx dotnet tracing",
         InstrumentedImageName,
-        new KeyValuePair<string, string>[]
+        new EnvVar[]
         {
             new("SIGNALFX_PROFILER_ENABLED", "0")
         });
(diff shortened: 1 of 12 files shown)


overhead-test/src/Configs/AgentConfig.cs
```diff
@@ -1,26 +1,58 @@
...
         DockerImageName = dockerImageName ?? throw new ArgumentNullException(nameof(dockerImageName));
-        ProfilerEnabled = profilerEnabled ?? throw new ArgumentNullException(nameof(profilerEnabled));
+        AdditionalEnvVars = additionalEnvVars ?? throw new ArgumentNullException(nameof(additionalEnvVars));
     }
 
     public string Name { get; }
+    public string Description { get; }
     public string DockerImageName { get; }
-    public string ProfilerEnabled { get; }
+    public IEnumerable<KeyValuePair<string, string>> AdditionalEnvVars { get; }
```
pellared, line 57: nit refactoring proposal:

```csharp
    public EnvVar[] AdditionalEnvVars { get; }

internal struct EnvVar
{
    public string Name;
    public string Value;
}
```

it should reduce a lot of noise from the code and having a type like `EnvVar` would be more meaningful than `KeyValuePair<string, string>`

overhead-test/src/Configs/TestConfig.cs
```diff
@@ -9,7 +9,13 @@ internal class TestConfig
...
         return new TestConfig(
-            new[] { AgentConfig.Baseline, AgentConfig.Instrumented, AgentConfig.Profiled },
+            new[]
+            {
+                AgentConfig.Baseline,
+                AgentConfig.Instrumented,
+                AgentConfig.Profiled,
+                AgentConfig.ProfiledHighFrequency
+            },
             1000,
```
lachmatt, line 19: I will increase number of iterations, once we have stable solution.

Kielek (approved):

PR #538, opened 2022-06-22
 Context propagators log errors and hit exceptions on typical case

dszmigielski:
## Why

When incoming requests has no context propagated, the context propagators log errors and hit exceptions.

## What

Fix the bug described above. Also improve the code of B3 propagator TryExtract method.

## Tests

CI passing
Tested manually by @pjanotti

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,13 +9,13 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
-- Fix context propagation on .NET Framework.
-
 ### General
 
 ### Breaking changes
 
 ### Bugfixes
+
+- Fix context propagation on .NET Framework.
 
 ### Enhancements
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
+- Fix context propagation on .NET Framework.
+
 ### General
 
 ### Breaking changes
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
@@ -56,7 +56,7 @@ namespace Datadog.Trace.ClrProfiler
                 return spanContext;
             }
 
-            return SpanContextPropagator.Instance.Extract(value, (headers, key) => new List<string> { headers[key] });
+            return SpanContextPropagator.Instance.Extract(value);
         }
 
         void IDistributedTracer.SetSpanContext(IReadOnlyDictionary<string, string> value)
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
@@ -3,8 +3,6 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-// Modified by Splunk Inc.
-
 using System;
 using System.Collections.Generic;
 using System.Threading;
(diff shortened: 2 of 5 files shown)


pellared:
When I quickly look at it it looks like some bug in another place 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
```diff
@@ -56,7 +56,7 @@ SpanContext IDistributedTracer.GetSpanContext()
                 return spanContext;
             }
 
-            return SpanContextPropagator.Instance.Extract(value, (headers, key) => new List<string> { headers[key] });
+            return SpanContextPropagator.Instance.Extract(value);
```
pellared, line 59: is this file now matching https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs and we can remove the "modified by SFx" comment?

tracer/src/Datadog.Trace/ClrProfiler/ManualTracer.cs
```diff
@@ -72,10 +72,8 @@ SpanContext IDistributedTracer.GetSpanContext()
             {
                 return spanContext;
             }
-            else
-            {
-                return SpanContextPropagator.Instance.Extract(values, (headers, key) => new List<string> { headers[key] });
-            }
+
+            return SpanContextPropagator.Instance.Extract(values);
```
pellared, line 76: I think you can refactor the file to matchhttps://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/ClrProfiler/ManualTracer.cs  and we can remove the "modified by SFx" comment

tracer/src/Datadog.Trace/Propagators/ParseUtility.cs
```diff
@@ -134,6 +134,11 @@ static bool TryParse(IEnumerable<string?> headerValues, ref bool hasValue, out u
             static bool TryParse(IEnumerable<string?> headerValues, ref bool hasValue, out int result)
             {
                 result = 0;
+                if (headerValues == null)
```
pellared, line 137: I am curious if this is really needed or is just in case. I think the passed argument is always at least `Enumerable.Empty<string>()`.

tracer/src/Datadog.Trace/Propagators/ParseUtility.cs
```diff
@@ -211,6 +216,11 @@ static bool TryParse(IEnumerable<string?> headerValues, ref bool hasValue, out i
             // IEnumerable version (different method to avoid try/finally in the caller)
             static string? ParseStringIEnumerable(IEnumerable<string?> headerValues)
             {
+                if (headerValues == null)
```
pellared, line 219: Same comment as before

tracer/src/Datadog.Trace/ClrProfiler/AutomaticTracer.cs
dszmigielski: Yes, that is correct - I'll change that

tracer/src/Datadog.Trace/Propagators/ParseUtility.cs
dszmigielski: It is not. If you try to remove it, it will stop working correctly for some cases (e.g. WebRequestTests using net461). I think we can investigate that further, but this is not main focus of this PR.

pellared:
I propose to add a line in `CHANGELOG.md`. E.g.

```
- Fix context propagation on .NET Framework.
```

tracer/src/Datadog.Trace/Propagators/ParseUtility.cs
pellared: I suggest investigating. I am curious why upstream does not need it and their tests are passing.

dszmigielski: I will create the task for that and start working on it right away.

pellared (approved):

docs/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this repository adheres to [Semantic Versioning](https://semver.org/spec/v2.
 
 ## Unreleased
 
+- Fix context propagation on .NET Framework.
```
pellared, line 12: This in bad sections 😄

pellared: https://github.com/signalfx/signalfx-dotnet-tracing/pull/538/commits/137754e2600a892fa28d8d9d5fc7b7b2e1bbe30e

PR #556, opened 2022-07-12
[AlwaysOnProfiling] Span context propagation fix

lachmatt:
## Why

Fix span context propagation.

## What

- convert to big-endian order when converting `long` to `bytes[]`
- extract exporting logic from `ThreadSampleExporter` to `OtlpHttpLogSender` for better testability

## Tests

Included in PR.

### DIFF
--- a/tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
+++ b/tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Net;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Propagation;
+using Datadog.Tracer.OpenTelemetry.Proto.Logs.V1;
+
+namespace Datadog.Trace.AlwaysOnProfiler;
+
+/// <summary>
+/// Sends logs in binary-encoded protobuf format over http.
+/// </summary>
+internal class OtlpHttpLogSender : ILogSender
+{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(OtlpHttpLogSender));
+
+    private readonly ImmutableExporterSettings _exporterSettings;
+
+    public OtlpHttpLogSender(ImmutableExporterSettings exporterSettings)
+    {
+        _exporterSettings = exporterSettings ?? throw new ArgumentNullException(nameof(exporterSettings));
+    }
+
+    public void Send(LogsData logsData)
+    {
+        HttpWebRequest httpWebRequest;
+
+        try
+        {
+            httpWebRequest = WebRequest.CreateHttp(_exporterSettings.LogsEndpointUrl);
+            httpWebRequest.ContentType = "application/x-protobuf";
+            httpWebRequest.Method = "POST";
+            httpWebRequest.Headers.Add(CommonHttpHeaderNames.TracingEnabled, "false");
+
+            using var stream = httpWebRequest.GetRequestStream();
+            Vendors.ProtoBuf.Serializer.Serialize(stream, logsData);
+            stream.Flush();
+        }
+        catch (Exception ex)
+        {
+            Log.Error("Exception preparing request to send thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, ex);
+            return;
+        }
+
+        try
+        {
+            using var httpWebResponse = (HttpWebResponse)httpWebRequest.GetResponse();
+
+            if (httpWebResponse.StatusCode >= HttpStatusCode.OK && httpWebResponse.StatusCode < HttpStatusCode.MultipleChoices)
+            {
+                return;
+            }
+
+            Log.Warning("HTTP error sending thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, httpWebResponse.StatusCode);
+        }
+        catch (Exception ex)
+        {
+            Log.Error("Exception sending thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, ex.Message);
+        }
+    }
+}
--- a/tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
+++ b/tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
@@ -3,22 +3,19 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Net;
 using Datadog.Trace.Configuration;
-using Datadog.Trace.ExtensionMethods;
-using Datadog.Trace.Logging;
-using Datadog.Trace.Propagation;
 using Datadog.Tracer.OpenTelemetry.Proto.Common.V1;
 using Datadog.Tracer.OpenTelemetry.Proto.Logs.V1;
 using Datadog.Tracer.OpenTelemetry.Proto.Resource.V1;
+using BitConverter = System.BitConverter;
 
 namespace Datadog.Trace.AlwaysOnProfiler
 {
     internal abstract class ThreadSampleExporter
     {
-        protected static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ThreadSampleExporter));
+        private readonly ILogSender _logSender;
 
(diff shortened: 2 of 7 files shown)


tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Net;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Propagation;
+using Datadog.Tracer.OpenTelemetry.Proto.Logs.V1;
+
+namespace Datadog.Trace.AlwaysOnProfiler;
+
+/// <summary>
+/// Sends logs in binary-encoded protobuf format over http.
```
pellared, line 11: nit

```suggestion
/// Sends logs in binary-encoded protobuf format over HTTP.
```

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
```diff
@@ -0,0 +1,61 @@
...
+/// </summary>
+internal class OtlpHttpLogSender : ILogSender
+{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(OtlpHttpLogSender));
+
+    private readonly ImmutableExporterSettings _exporterSettings;
+
+    public OtlpHttpLogSender(ImmutableExporterSettings exporterSettings)
```
pellared, line 19: What do you think of depending only on `string endpointUrl` instead? The caller would simply do `_exporterSettings.LogsEndpointUrl`.

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
```diff
@@ -0,0 +1,61 @@
...
+            httpWebRequest.Method = "POST";
+            httpWebRequest.Headers.Add(CommonHttpHeaderNames.TracingEnabled, "false");
+
+            using var stream = httpWebRequest.GetRequestStream();
+            Vendors.ProtoBuf.Serializer.Serialize(stream, logsData);
+            stream.Flush();
+        }
+        catch (Exception ex)
+        {
+            Log.Error("Exception preparing request to send thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, ex);
```
dszmigielski, line 41: Is there a reason not to use `Log.Error(exception, message)` overload?

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
```diff
@@ -0,0 +1,61 @@
...
+            if (httpWebResponse.StatusCode >= HttpStatusCode.OK && httpWebResponse.StatusCode < HttpStatusCode.MultipleChoices)
+            {
+                return;
+            }
+
+            Log.Warning("HTTP error sending thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, httpWebResponse.StatusCode);
+        }
+        catch (Exception ex)
+        {
+            Log.Error("Exception sending thread samples to {0}: {1}", _exporterSettings.LogsEndpointUrl, ex.Message);
```
dszmigielski, line 58: Same as above

tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
```diff
@@ -28,16 +25,13 @@ protected ThreadSampleExporter(ImmutableTracerSettings tracerSettings)
...
+            _logSender = logSender ?? throw new ArgumentNullException(nameof(logSender));
         }
 
-        protected ReadOnlyCollection<KeyValue> FixedLogRecordAttributes { get; }
-
-        protected Uri LogsEndpointUrl { get; }
+        private ReadOnlyCollection<KeyValue> FixedLogRecordAttributes { get; }
 
-        protected LogsData LogsData { get; }
+        private LogsData LogsData { get; }
```
dszmigielski, line 34: It can be a readonly private field now, I think

tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
```diff
@@ -58,63 +52,51 @@ public void ExportThreadSamples(List<ThreadSample> threadSamples)
...
+                // The exporter reuses the LogsData object, but the actual log records are not
                 // needed after serialization, release the log records so they can be garbage collected.
                 LogsData.ResourceLogs[0].InstrumentationLibraryLogs[0].Logs.Clear();
             }
+        }
 
-            try
-            {
-                using var httpWebResponse = (HttpWebResponse)httpWebRequest.GetResponse();
+        private static byte[] GetBigEndianBytes(long high, long low)
```
pellared, line 74: nit: how about renaming the methods to `ToBigEndianBytes`?

pellared (approved):
nice 👍 I just left a few nit comments

tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
dszmigielski, line 74: Just an idea: is this method bound to the Exporter class, or can we extract it to somewhere else for cleaner code and future usage?

dszmigielski (approved):
Added some comments, but no blockers from my side

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/556/commits/dc6fc6e97d531e4aed6333e582fdaa7aafd6c6c9

tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/556/commits/dc6fc6e97d531e4aed6333e582fdaa7aafd6c6c9

pellared: Personally, I prefer to have it here if it is the only place where it is used 😉

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
lachmatt: I don't think there is, it was extracted as is and I haven't noticed

tracer/src/Datadog.Trace/AlwaysOnProfiler/ThreadSampleExporter.cs
lachmatt: I agree with @pellared and prefer to leave it as is

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/556/commits/4ca88704a8fc6ecee854bd1aea6251f6e30cde43

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/556/commits/4ca88704a8fc6ecee854bd1aea6251f6e30cde43

tracer/src/Datadog.Trace/AlwaysOnProfiler/OtlpHttpLogSender.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/556/commits/4ca88704a8fc6ecee854bd1aea6251f6e30cde43

pellared (approved):

PR #560, opened 2022-07-14
[AlwaysOn Profiling] fix span context propagation in pprof exporter

lachmatt:
## Why

Fix span context propagation for `Pprof` exporter.

## What
- set `span_id` label value as a `hex` string
- set span context info on `LogRecord` only for `PlainTextThreadSampleExporter`

## Tests

Included in CI.
Stack samples with span context properly associated:
`Text` exporter: https://app.signalfx.com/#/apm/traces/39a34bba0517d79a24db86a972ec4231
`Pprof` exporter: https://app.signalfx.com/#/apm/traces/54362b8fa82f868f3033d3fee1bda6dd

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/AlwaysOnProfiler/PprofThreadSampleExporterTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/AlwaysOnProfiler/PprofThreadSampleExporterTests.cs
@@ -1,0 +1,81 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using Datadog.Trace.AlwaysOnProfiler;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Vendors.ProtoBuf;
+using Datadog.Tracer.Pprof.Proto.Profile;
+using Xunit;
+
+namespace Datadog.Trace.Tests.AlwaysOnProfiler;
+
+public class PprofThreadSampleExporterTests
+{
+    [Fact]
+    public void If_stack_sample_has_span_context_associated_then_it_is_sent_inside_labels()
+    {
+        var sender = new TestSender();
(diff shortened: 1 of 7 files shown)


tracer/test/Datadog.Trace.Tests/AlwaysOnProfiler/PprofThreadSampleExporterTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        var spanId = GetId("span_id", profile);
+
+        // hex representation of 1234567890
+        Assert.Equal("00000000499602d2", spanId);
+
+        var traceId = GetId("trace_id", profile);
+
+        // concatenated hex representations of 1234567890 and 9876543210
+        Assert.Equal("00000000499602d2000000024cb016ea", traceId);
```
pellared, line 51: if you want all assertions to be run (even if one will fail) you can use FluentAssertion + AssertionScope

tracer/test/Datadog.Trace.Tests/AlwaysOnProfiler/PprofThreadSampleExporterTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                TraceIdHigh = 1234567890,
+                Timestamp = new ThreadSample.Time(10000),
+            }
+        });
+
+        var log = sender.SentLogs[0];
+
+        // for Pprof exporter, span context SHOULD NOT be set on logRecord level
+        Assert.Null(log.TraceId);
+        Assert.Null(log.SpanId);
```
pellared, line 39: I know that it was a bug, but I would remove it. It was for sure handy during development but I do not see a lot of value in it after we have the fix 😉

pellared (approved):

pjanotti (approved):

aurbiztondo-splunk (approved):

PR #564, opened 2022-07-26
Release 0.2.7.0

dszmigielski:
### Bugfixes

- Fix span context propagation for AlwaysOn Profiling

### DIFF
--- a/tracer/tools/PipelineMonitor/PipelineMonitor.csproj
+++ b/tracer/tools/PipelineMonitor/PipelineMonitor.csproj
@@ -8,7 +8,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="Datadog.Trace" Version="0.2.6" />
+      <PackageReference Include="Datadog.Trace" Version="0.2.7" />
     </ItemGroup>
 
 </Project>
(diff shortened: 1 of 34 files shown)


aurbiztondo-splunk (approved):

tracer/tools/PipelineMonitor/PipelineMonitor.csproj
```diff
@@ -8,7 +8,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="Datadog.Trace" Version="0.2.6" />
+      <PackageReference Include="Datadog.Trace" Version="0.2.7" />
```
pjanotti, line 11: This is something that we don't want to update.

pjanotti (commented):
Actually LGTM minus the nuget package that we don't care about setting to v0.2.7.0.

tracer/tools/PipelineMonitor/PipelineMonitor.csproj
pjanotti: Please, check if the tool can be changed to avoid this change.

Kielek (approved):

PR #566, opened 2022-07-29
[AlwaysOn Profiling] Don't set timestamp on LogRecord for pprof

lachmatt:
## Why

Adjust to updated spec.

## What

Don't set `timestamp` on `LogRecord` for `pprof` exporter.

## Tests

Included in PR.


Kielek (approved):

pjanotti (commented):
LGTM

PR #568, opened 2022-08-04
Do not mark 404 responses as errors

Kielek:
## Why

SWAT ticket.

## What

There is possibility to throw exception which is handled late on server side and converted to 404 instead of 500.
This PR fixes such cases.

## Tests

New test cases created. For differences please check cbd2e42042aa9a2467e182589dc2f02b2ece16ef and d069f5fd2d97115eb52a24d20f743c3cb64f499b

### DIFF
--- a/Datadog.Trace.sln
+++ b/Datadog.Trace.sln
@@ -436,7 +436,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.Telemetry", "tracer
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Sandbox.AutomaticInstrumentation", "tracer\test\test-applications\regression\Sandbox.AutomaticInstrumentation\Sandbox.AutomaticInstrumentation.csproj", "{10619BA2-AED1-482A-8570-BB7C7B83DDDC}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Datadog.Trace.Coverage.collector", "tracer\src\Datadog.Trace.Coverage.collector\Datadog.Trace.Coverage.collector.csproj", "{7A0D8A39-C60D-4BA1-93AB-D429080BBFA5}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.Trace.Coverage.collector", "tracer\src\Datadog.Trace.Coverage.collector\Datadog.Trace.Coverage.collector.csproj", "{7A0D8A39-C60D-4BA1-93AB-D429080BBFA5}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.TraceAnnotations", "tracer\test\test-applications\integrations\Samples.TraceAnnotations\Samples.TraceAnnotations.csproj", "{230A9B80-E1AA-469F-86F2-B7E257F14E66}"
 EndProject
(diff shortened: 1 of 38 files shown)


dszmigielski (approved):

Kielek:
[windows-msi-x64.zip](https://github.com/signalfx/signalfx-dotnet-tracing/files/9260835/windows-msi-x64.zip)

Datadog.Trace.sln
```diff
@@ -468,16 +468,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.InstrumentedAssembl
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.InstrumentedAssemblyVerification.Standalone", "tracer\src\Datadog.InstrumentedAssemblyVerification.Standalone\Datadog.InstrumentedAssemblyVerification.Standalone.csproj", "{61FD6A7E-38EB-4B90-A890-DE3205C041B1}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.AspNetMvc5CustomException", "tracer\test\test-applications\aspnet\Samples.AspNetMvc5CustomException\Samples.AspNetMvc5CustomException.csproj", "{F71B1B2D-C3E0-49B6-AC29-B96F45E12D5E}"
```
pjanotti, line 471: nit: per https://www.codeproject.com/Reference/720512/List-of-Visual-Studio-Project-Type-GUIDs this may not be the most appropriate GUID for this project. Was it added by VS or `dotnet sln`?

tracer/test/test-applications/aspnet/Samples.AspNetMvc5CustomException/Global.asax.cs
```diff
@@ -0,0 +1,20 @@
+// Modified by Splunk Inc.
```
pjanotti, line 1: The other files added for this don't have this.

tracer/src/Datadog.Trace/AspNet/TracingHttpModule.cs
```diff
@@ -210,6 +210,11 @@ private void OnEndRequest(object sender, EventArgs eventArgs)
                             scope.Span.ResourceName = $"{app.Request.HttpMethod.ToUpperInvariant()} {path.ToLowerInvariant()}";
                         }
 
+                        if (app.Context.Items[SharedItems.HttpContextPropagatedExceptionKey] is Exception exception && app.Context.Response.StatusCode != 404)
+                        {
+                            scope.Span.SetException(exception);
```
pjanotti, line 215: I wonder if there is any case that one may want to record the exception but not mark the span as in error.

pjanotti (approved):
LGTM. The comments are minor.

tracer/src/Datadog.Trace/AspNet/TracingHttpModule.cs
pjanotti: Let's consider that separately.

Datadog.Trace.sln
Kielek: By VS. The same GUID is used for Samples.AspNetMvc5. I would keep it.

tracer/test/test-applications/aspnet/Samples.AspNetMvc5CustomException/Global.asax.cs
Kielek: Added to other c# files

tracer/test/test-applications/aspnet/Samples.AspNetMvc5CustomException/Handlers/CustomNofFoundExceptionMessageHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Samples.AspNetMvc5CustomException.Handlers
+{
+    public class CustomNofFoundExceptionMessageHandler : DelegatingHandler
```
lachmatt, line 11: nit: typo in class name

Kielek: fixed

lachmatt (approved):

pjanotti (approved):

PR #585, opened 2022-08-18
Fix sampled value in ServerTiming header

pjanotti:
## Why

ServerTiming is not taking into account the TraceContext and can set sampled to "00"

## What

Consider TraceContext and as a fallback, if no sampling priority is available, fallback to sampled.

## Tests

Added the case of SamplingPriority null to ensure that the fallback is working as intended.


tracer/src/Datadog.Trace/ServerTimingHeader.cs
```diff
@@ -37,7 +36,9 @@ public static void SetHeaders<T>(SpanContext context, T carrier, Action<T, strin
 
         private static string ToHeaderValue(SpanContext context)
         {
-            return string.Format(ServerTimingFormat, W3CContextPropagator.GetFormattedTraceParent(context));
+            var samplingPriority = context.TraceContext?.SamplingPriority ?? context.SamplingPriority ?? SamplingPriorityValues.AutoKeep;
```
pjanotti, line 39: This is a bit different than the propagators. On the propagators there is no fallback to `SamplingPriorityValues.AutoKeep`. It doesn't seem to be possible to reach that in practice but it is also hard to be 100% sure about it. The reason to have this fallback is that if we are adding the ServerTiming header the assumption is that is being sampled and it seems possible to create the span and set the sampling priority later. This way it preserves the propagators behavior regarding Inject/Extract and ensures that ServerTiming is sampled is set even if the context still is going to have its value set later.

dszmigielski (approved):

aurbiztondo-splunk (approved):

dszmigielski (approved):

Kielek (approved):

PR #586, opened 2022-08-19
remove agent word from overhead tests

Kielek:
## Why

We are avoiding callling .NET instrumentation as agents.

## What

remove agent word from overhead tests

## Tests

CI


lachmatt (approved):

pjanotti (commented):
Good catch.

PR #599, opened 2022-08-30
Allocation sampling prototype, with text printout

johnbley:
Barebones text printouts of allocation samples, controlled with an environment variable.  This approach (AllocationTick events) was settled on after a series of prototypes showing this as having the lowest overhead.


tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -266,6 +266,27 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
+            if (this->threadSampler == nullptr)
+            {
+                // FIXME may want to combine precheck logic to ensure only one allocation site
+                this->threadSampler = new always_on_profiler::ThreadSampler();
+            }
+            this->threadSampler->StartAllocationSampling(info12);
+        }
+        else
+        {
+            Logger::Error("Could not enable allocation sampling: CLR version not supported");
```
pjanotti, line 284: It would be nice to have the `hr` on this message.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -624,6 +624,31 @@ void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
 #endif
 }
 
+void ThreadSampler::AllocationTick(ULONG dataLen, LPCBYTE data)
+{
+    // FIXME find a symbolic way into this rather than byte offsets
+    uint64_t allocatedSize = *((uint64_t*) &(data[dataLen - 8]));
+#ifdef _WIN32
```
pjanotti, line 631: What is different in WIN32?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -624,6 +624,31 @@ void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
...
+    printf("Allocation: %i %ws\n", (int) allocatedSize, (wchar_t*) &data[26]);
```
pjanotti, line 632: I suppose the string here is the object type of the allocation triggering the event. 

Even for initial experimentation it is a nice to have the threadID and the heap id also displayed, if not too cumbersome to add right now.

pjanotti (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -624,6 +624,31 @@ void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
 #endif
 }
 
+void ThreadSampler::AllocationTick(ULONG dataLen, LPCBYTE data)
+{
+    // FIXME find a symbolic way into this rather than byte offsets
```
Kielek, line 629: Could you please create jira items for all FIXMEs (we need visibility for the progress)?
What is more we have a convention to mark it as `// TODO Splunk: `. Just to have easy way to find it.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -624,6 +624,31 @@ void ThreadSampler::StartSampling(ICorProfilerInfo10* cor_profiler_info10)
...
+    printf("Allocation: %i %ws\n", (int) allocatedSize, (wchar_t*) &data[26]);
+#else
+    printf("Allocation: %i\n", (int) allocatedSize);
+#endif
+}
+
+void ThreadSampler::StartAllocationSampling(ICorProfilerInfo12* info12)
+{
+    EVENTPIPE_SESSION session;
+    COR_PRF_EVENTPIPE_PROVIDER_CONFIG sessionConfig[] = {{WStr("Microsoft-Windows-DotNETRuntime"),
```
Kielek, line 641: I see here a Windows in the configuration. Is it intended for Linux and MacOS?
I suppose that it is something on .NET side itself, but I would like to double check.

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -266,6 +266,27 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
     }
+    if (IsAllocationSamplingEnabled())
+    {
+        ICorProfilerInfo12 * info12 = nullptr;
+        HRESULT hr = info_->QueryInterface(__uuidof(ICorProfilerInfo12), (void**) &info12);
+        if (!FAILED(hr) && info12 != nullptr)
+        {
+            if (this->threadSampler == nullptr)
+            {
+                // FIXME may want to combine precheck logic to ensure only one allocation site
```
Kielek, line 277: New jira item.

tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
```diff
@@ -112,6 +112,7 @@ namespace environment {
 
     // If you change this, change corresponding logic in Instrument.cs too
     const shared::WSTRING thread_sampling_enabled = WStr("SIGNALFX_PROFILER_ENABLED");
+    const shared::WSTRING allocation_sampling_enabled = WStr("SIGNALFX_ALLOCATION_PROFILER_ENABLED");
```
Kielek, line 115: Please change at least the env. var to `SIGNALFX_PROFILER_MEMORY_ENABLED`.
Consider if the const name is fine in this context.

ref: https://github.com/signalfx/gdi-specification/blob/41ab06528a51b7d0a7f10a2b17e22ec0e88c4324/specification/configuration.md?plain=1#L168

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.h
```diff
@@ -82,6 +82,8 @@ class ThreadSampler
 {
```
Kielek, line 82: You are adding new feature to the purely thread/cpu sampler.

Please try to find better name for this. (It will probably propagate for other places in the code.
It could be separate PR/new jira item.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
johnbley: Don't have a running linux env to test on at the moment, so just stubbing some stuff out so it will compile.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
johnbley: When I get my linux env back up and running I will manually test and tweak that; added a TODO for the moment.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
johnbley: Switched to TODO, will mark jira in one swoop after a few rounds of PRs on this.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.h
johnbley: I agree that a better name for `ThreadSampler` would probably be `AlwaysOnProfiler` at this point but the refactoring can wait until we figure out (e.g.) where and how name cache state is shared.

Kielek (approved):

PR #600, opened 2022-08-31
CPU profiling reached GA

Kielek:
## Why

CPU profiling reached GA

## What

removing experimental statement

## Tests

N/A

### DIFF
--- a/docs/always-on-profiling.md
+++ b/docs/always-on-profiling.md
@@ -1,9 +1,6 @@
 
 # About the .NET AlwaysOn Profiling
 
-> :construction: &nbsp;Status: Experimental - exported telemetry and
-> configuration settings may change.
-
 The SignalFx Instrumentation for .NET includes a continuous thread sampler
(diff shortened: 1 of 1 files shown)


pjanotti (approved):

pellared (approved):
The link 

> Please check [description](https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/internal-config.md) for the following environment variables

is invalid. Worth to fix.

Also, I would rather reference `advanced-config.md#alwayson-profiling-settings` than duplicate the same description.

Kielek:
> The link
> 
> > Please check [description](https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/internal-config.md) for the following environment variables
> 
> is invalid. Worth to fix.
> 
> Also, I would rather reference `advanced-config.md#alwayson-profiling-settings` than duplicate the same description.

Link fixed.  I would say that current version is good enough.

dszmigielski (approved):

PR #606, opened 2022-09-05
[RuntimeMetrics] Reimplement to match otel - gc

lachmatt:
## Why

Adjust metrics implementation to match [`otel`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs).

## What

- reimplement gc-related metrics, threadpool's thread count, contention 
- adjust metrics export

## Tests

Included in PR.
Dashboard: https://app.signalfx.com/#/dashboard/FGfk_uBAgAA?groupId=FGfk7HwAcAE&configId=FGflAd_AcAA&startTimeUTC=1662357568000&endTimeUTC=1662358037000

### DIFF
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/AzureAppServicePerformanceCounters.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/AzureAppServicePerformanceCounters.cs
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
+// Modified by Splunk Inc.
+
 #if NETFRAMEWORK
 
 using System;
@@ -16,14 +18,10 @@ namespace Datadog.Trace.RuntimeMetrics
     internal class AzureAppServicePerformanceCounters : IRuntimeMetricsListener
     {
         internal const string EnvironmentVariableName = "WEBSITE_COUNTERS_CLR";
-        private const string GarbageCollectionMetrics = $"{MetricsNames.Gen0HeapSize}, {MetricsNames.Gen1HeapSize}, {MetricsNames.Gen2HeapSize}, {MetricsNames.LohSize}, {MetricsNames.Gen0CollectionsCount}, {MetricsNames.Gen1CollectionsCount}, {MetricsNames.Gen2CollectionsCount}";
+        private const string GarbageCollectionMetrics = $"{MetricsNames.Gc.HeapSize}, {MetricsNames.Gc.CollectionsCount}";
 
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<AzureAppServicePerformanceCounters>();
         private readonly IDogStatsd _statsd;
-
-        private int? _previousGen0Count;
-        private int? _previousGen1Count;
-        private int? _previousGen2Count;
 
         public AzureAppServicePerformanceCounters(IDogStatsd statsd)
         {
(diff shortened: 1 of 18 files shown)


dszmigielski (approved):

tracer/src/Datadog.Trace/RuntimeMetrics/GcMetrics.cs
```diff
@@ -0,0 +1,36 @@
+// Modified by Splunk Inc.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Datadog.Trace.Vendors.StatsdClient;
+
+namespace Datadog.Trace.RuntimeMetrics;
+
+// source originated from: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```
Kielek, line 10: Please use perm-link. This can be changed in the future.

Kielek (approved):

tracer/src/Datadog.Trace/RuntimeMetrics/GcMetrics.cs
```diff
@@ -0,0 +1,36 @@
...
+    public static void PushCollectionCounts(IDogStatsd dogStatsd)
+    {
+        long collectionsFromHigherGeneration = 0;
+
+        for (var gen = NumberOfGenerations - 1; gen >= 0; --gen)
+        {
+            long collectionsFromThisGeneration = GC.CollectionCount(gen);
+            var currentValue = collectionsFromThisGeneration - collectionsFromHigherGeneration;
+
+            dogStatsd.Counter(MetricsNames.Gc.CollectionsCount, currentValue, tags: new[] { GenerationTag(gen) });
```
pjanotti, line 31: Can we cache these arrays instead of creating a new one on every collection?

pjanotti (approved):
LGTM. Just minor suggestions.

tracer/src/Datadog.Trace/RuntimeMetrics/AzureAppServicePerformanceCounters.cs
```diff
@@ -39,33 +37,12 @@ public void Refresh()
...
-                _statsd.Increment(MetricsNames.Gen2CollectionsCount, gen2 - _previousGen2Count.Value);
-            }
-
-            _previousGen0Count = gen0;
-            _previousGen1Count = gen1;
-            _previousGen2Count = gen2;
+            _statsd.Gauge(MetricsNames.Gc.HeapSize, value.Gen0Size, tags: new[] { GcMetrics.GenerationTag(0) });
+            _statsd.Gauge(MetricsNames.Gc.HeapSize, value.Gen1Size, tags: new[] { GcMetrics.GenerationTag(1) });
+            _statsd.Gauge(MetricsNames.Gc.HeapSize, value.Gen2Size, tags: new[] { GcMetrics.GenerationTag(2) });
+            _statsd.Gauge(MetricsNames.Gc.HeapSize, value.LohSize, tags: new[] { GcMetrics.GenerationTag(3) });
```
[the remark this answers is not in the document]
lachmatt: > That seems confuse to me: very few people refer to the LOH as generation 3. That said if OTel is doing it that way we want to follow it here.

Tag generated by this invocation is `"generation:loh"`. Parameter is used as an index into `GenNames` (name comes from [`otel`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/bc947a00c3f859cc436f050e81172fc1f8bc09d7/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs#L45))


> Below there is a metric for POH shouldn't we have a gauge for it too?

This code is specific to dotnet framework.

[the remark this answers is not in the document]
pjanotti: I understood the code after I got to the end of the review and put the connections that you described in my mind. I think things would be easier to read if instead of `GcMetrics.GenerationTag(3)` we had something like `GcMetrics.GenerationTag(GcGenerationTag.LargeHeapObject)` or something better worded than that.

tracer/src/Datadog.Trace/RuntimeMetrics/GcMetrics.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/611

tracer/src/Datadog.Trace/RuntimeMetrics/AzureAppServicePerformanceCounters.cs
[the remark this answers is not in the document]
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/611

tracer/src/Datadog.Trace/RuntimeMetrics/GcMetrics.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/611

PR #617, opened 2022-09-14
Capture call stacks in allocation samples

johnbley:
Add a lock around usage of the `SamplingHelper` name cache(s).  Clear the volatile cache when capturing an allocation sample call stack.  Share out the code for parsing call stacks in C# to read this new data.  The `FrameCallback` (that can only take one parameter) now takes a new type to distinguish that we want to use the global name cache but an AllocationTick-local buffer to write data into, while the original CPU sampler wants something slightly different.  This (separating name cache from buffer management) will be refactored in a followup PR rather than polluting the locking changes further in this one.


lachmatt (approved):

Kielek (approved):

Kielek:
@johnbley, code looks good, but there is some compilation issues on Linux.

PR #621, opened 2022-09-19
Allocation subsampling algorithm

johnbley:
Introduce an algorithm for subsampling of allocations, with unit tests.  Algorithm described in more detail in comments, but note that it isn't wired up to anything yet since we should have a discussion about the constants involved and/or configuration of them.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
@@ -752,11 +752,11 @@ void CaptureAllocationStack(AlwaysOnProfiler* prof, ThreadSamplesBuffer* buffer)
     }
 }
 
-AllocationSubSampler::AllocationSubSampler(uint32_t targetPC, uint32_t secondsPC) :
-    targetPerCycle(targetPC), secondsPerCycle(secondsPC), seenThisCycle(0), sampledThisCycle(0), seenLastCycle(0),
+AllocationSubSampler::AllocationSubSampler(uint32_t targetPerCycle_, uint32_t secondsPerCycle_) :
+    targetPerCycle(targetPerCycle_), secondsPerCycle(secondsPerCycle_), seenThisCycle(0), sampledThisCycle(0), seenLastCycle(0),
     nextCycleStartMillis(
         std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())),
-    sampleLock(), rand()
+    sampleLock(), rand(std::default_random_engine((unsigned int)(nextCycleStartMillis.count())))
 {
 }
 void AllocationSubSampler::AdvanceCycle(std::chrono::milliseconds now)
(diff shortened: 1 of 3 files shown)


pjanotti (approved):
LGTM. One thing in my mind is that we may have a lot of contention on that lock when using server GC. That said seems reasonable to start with something correct and simple and optimize as needed.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -752,6 +752,57 @@ void CaptureAllocationStack(AlwaysOnProfiler* prof, ThreadSamplesBuffer* buffer)
     }
 }
 
+AllocationSubSampler::AllocationSubSampler(uint32_t targetPC, uint32_t secondsPC) :
```
Kielek, line 755: Different parameter names than in header file.

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -752,6 +752,57 @@ void CaptureAllocationStack(AlwaysOnProfiler* prof, ThreadSamplesBuffer* buffer)
...
+    targetPerCycle(targetPC), secondsPerCycle(secondsPC), seenThisCycle(0), sampledThisCycle(0), seenLastCycle(0),
+    nextCycleStartMillis(
+        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())),
+    sampleLock(), rand()
```
Kielek, line 759: Rand is always seed by default value - 1 - https://cplusplus.com/reference/random/linear_congruential_engine/linear_congruential_engine/
Is it expected?

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
```diff
@@ -752,6 +752,57 @@ void CaptureAllocationStack(AlwaysOnProfiler* prof, ThreadSamplesBuffer* buffer)
...
+}
+void AllocationSubSampler::AdvanceCycle(std::chrono::milliseconds now)
+{
+    nextCycleStartMillis = now + std::chrono::seconds(secondsPerCycle);
+    seenLastCycle = seenThisCycle;
+    seenThisCycle = 0;
+    sampledThisCycle = 0;
+}
+
+// We want to sample T items out of N per unit time, where N is unknown and may be < S or may be orders
```
Kielek, line 770: `S` is not defined here, same in the next line.

Kielek (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/always_on_profiler.cpp
johnbley: Whoops!  Assumed the default would be more sane than that; fixed to seed with time.

Kielek (approved):

johnbley:
> LGTM. One thing in my mind is that we may have a lot of contention on that lock when using server GC. That said seems reasonable to start with something correct and simple and optimize as needed.

Agreed, though I think in that scenario (heavily contended AllocationTicks) I think the buffer lock might be a bigger issue.

PR #626, opened 2022-09-28
Hook up allocation subsampling to the ticks

johnbley:
I chose an arbitrary default of (max) 200 allocation samples per minute.  Open to arguments over that or we can merge this and change later after more discussion.


tracer/src/Datadog.Trace.ClrProfiler.Native/environment_variables.h
```diff
@@ -115,6 +115,7 @@ namespace environment {
     const shared::WSTRING allocation_sampling_enabled = WStr("SIGNALFX_PROFILER_MEMORY_ENABLED");
 
     const shared::WSTRING thread_sampling_period = WStr("SIGNALFX_PROFILER_CALL_STACK_INTERVAL");
+    const shared::WSTRING max_allocation_samples_per_minute = WStr("SIGNALFX_PROFILER_MAX_MEMORY_SAMPLES_PER_MINUTE");
```
Kielek, line 118: Please mention it in internal md file.

Kielek (approved):

dszmigielski (approved):

PR #629, opened 2022-09-30
[AlwaysOnProfiler] Managed parser changes

lachmatt:
## Why

Prepare for exporter changes.

## What

Refactor parser.

## Tests

Included in PR.


tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleNativeFormatParser.cs
```diff
@@ -187,63 +131,153 @@ internal List<ThreadSample> Parse()
...
+                            NativeId = nativeId,
+                            ThreadName = threadName
+                        };
+
+                        var code = ReadShort(buffer, ref position);
+
+                        // each allocation sample has independently coded strings
+                        var codeDictionary = new Dictionary<int, string>();
+                        ReadStackFrames(code, threadSample, codeDictionary, buffer, ref position);
+                        allocationSamples.Add(new AllocationSample(new AllocationDetails(allocatedSize, typeName), threadSample));
```
dszmigielski, line 179: I would get rid of `AllocationDetails` class. It is used only here, holds 2 properties that could be held directly in `AllocationSample`, and makes this line look weird :)

dszmigielski (approved):

tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleNativeFormatParser.cs
Kielek: and it will be one less class on the heap :)

Kielek (approved):

dszmigielski (approved):

Kielek (approved):

tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleNativeFormatParser.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/629/commits/3d774407e4d7b4e7ae5a98be661a10fad8c963c1

PR #632, opened 2022-10-05
[AlwaysOnProfiler] Export allocation samples

lachmatt:
## Why

Export captured allocation samples.

## What

Build `LogRecord`s as required by [spec](https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md).

## Tests

Included in PR.

### DIFF
--- a/tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleBuilder.cs
+++ b/tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleBuilder.cs
@@ -4,16 +4,23 @@ using System.Collections.Generic;
 using System.Linq;
 using Datadog.Tracer.Pprof.Proto.Profile;
 
-namespace Datadog.Trace.AlwaysOnProfiler.Builder
+namespace Datadog.Trace.AlwaysOnProfiler
 {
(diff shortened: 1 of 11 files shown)


tracer/src/Datadog.Trace/AlwaysOnProfiler/SampleBuilder.cs
```diff
@@ -4,19 +4,26 @@
...
 using Datadog.Tracer.Pprof.Proto.Profile;
 
-namespace Datadog.Trace.AlwaysOnProfiler.Builder
+namespace Datadog.Trace.AlwaysOnProfiler
 {
     internal class SampleBuilder
     {
         private readonly Sample _sample = new();
         private readonly IList<ulong> _locationIds = new List<ulong>();
+        private readonly IList<long> _values = new List<long>();
```
Kielek, line 13: Candidate for separate PR: try to optimize memory usage. Probably we can keep just one item.
Please verify also _locationIds.

Kielek (approved):

dszmigielski (approved):

pjanotti (approved):

PR #635, opened 2022-10-10
Add PowerShell install module

RassK:
## Why

Supports https://github.com/signalfx/splunk-otel-collector/pull/2068

## What

Adds PowerShell install script to the release artifacts.

(must be manually added to the current artifacts)


dszmigielski (approved):
LGTM

install.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
+
+# Download installer MSI to Temp
+$msi = Join-Path $env:temp $download.name
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
+
+# Install downloaded MSI
+Start-Process msiexec.exe -Wait -ArgumentList "/I $msi /quiet"
```
pellared, line 24: shouldn't we check the error code? 

https://learn.microsoft.com/en-us/windows/win32/msi/error-codes

install.ps1
```diff
@@ -0,0 +1,27 @@
...
+
+# determine OS architecture
+$os_bits = (Get-CimInstance Win32_OperatingSystem).OSArchitecture
+$os_arch = (&{If($os_bits -eq "64-bit") {"x64"} Else {"x86"}})
+
+# File pattern to search for
+$pattern = "signalfx-dotnet-tracing-*-$os_arch.msi"
+
+# Find latest MSI to download
+$download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
```
pellared, line 17: How this script will be used? How `Invoke-WebRequest` errors will be handled?

RassK: I guess the invoker should check for the errors, but we should be sure to propagate them, not to hide.

install.ps1
RassK: https://github.com/signalfx/signalfx-dotnet-tracing/pull/635/commits/220d32ab82a828d54fffa6c986281a9a253b1ef2

install.ps1
```diff
@@ -0,0 +1,32 @@
...
+
+# Download installer MSI to Temp
+$msi = Join-Path $env:temp $download.name
+Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
+
+# Install downloaded MSI
+$process = Start-Process msiexec.exe -Wait -PassThru -ArgumentList "/I $msi /quiet"
+
+if($process.ExitCode -ne 0 -or $process.ExitCode -ne 3010){
```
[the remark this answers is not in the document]
RassK: Ah... of course.

install.ps1
pellared: I am not sure if the caller can pass `-ErrorAction Stop` when running a such script.



We could probably

1. Ask the caller would call `$ErrorActionPreference = "Stop"` before using this script.
1. Remember the current action preference and then alter it to "Error" at the top.  Then put everything to `try` and in `finally` set the previous action preference.
2. Pass `-ErrorAction Stop` when invoking `Invoke-WebRequest` and other functions (I think it might be very error-prone)
3. Refactor the script to an advanced function. Then  `-ErrorAction Stop`  can be passed when invoking the script.

RassK: Just wondering when doing like this, the `$ErrorActionPreference` is default, but it still stops in the right place?

```powershell
try {
    Invoke-Expression "& $PSScriptRoot\install.ps1"
}
catch {
    "An Error occurred."
}
```

Kielek (approved):

install.ps1
pellared: Have you tried it? 😅 


The default value for `$ErrorActionPreference` is `Continue`

.github/workflows/release-draft.yml
```diff
@@ -61,7 +61,9 @@ jobs:
         uses: actions/upload-artifact@v3.1.0
         with:
           name: artifacts
-          path: tracer/bin/artifacts/*/en-us
+          path: |
+            tracer/bin/artifacts/*/en-us
+            install.ps1
```
pellared, line 66: Do you want to add it to the GH release? If so then you also need to also update the `create-release` job.

install.ps1
RassK: Yep, I just discovered that try \ catch is modifying how it's executing. Without try \ catch, it will fail and continue.

.github/workflows/release-draft.yml
RassK: Wondering if I can use `artifacts/install.ps1` or due ubuntu I should be  `artifacts/install.*`

install.ps1
pjanotti: @RassK so it still need to be changed to stop on first failure, right?

pjanotti (approved):
LGTM, although I'm unsure about current behavior in case of download error.

RassK:
> LGTM, although I'm unsure about current behavior in case of download error.

https://github.com/signalfx/signalfx-dotnet-tracing/pull/635#discussion_r991259231
If invoked like this, everything will be good.

Other way is to do it ourselves... or do like @pellared mentioned, configure `ErrorActionPreference` to `"stop"`, but stopping is not giving us cleanup time.

RassK:
Another idea, as SIG decided to convert OTel's script to more of a library, should we do the same?

pjanotti:
> Another idea, as SIG decided to convert OTel's script to more of a library, should we do the same?

It is a good idea. How do you prefer to handle it? Merge script first then convert or change directly in this PR?

install.ps1
RassK: Technically no, as the try block involves most of the logic... but now when converting to the module, the execution process will be a bit more complicated. So all failures can be handled internally in right order.

pellared (approved):

Kielek (approved):

PR #637, opened 2022-10-11
Add FOSSA scan

pjanotti:
## Why

To enable FOSSA scan

## What

Follow the receipt to add a FOSSA scan on gitlab, see https://splunk.atlassian.net/wiki/spaces/PROD/pages/1078131728850/How+to+run+FOSSA+Scans#Integrating-FOSSA-into-Gitlab-CI.

## Tests

N/A


Kielek (approved):

.gitlab-ci.yml
```diff
@@ -177,4 +181,10 @@ run-overhead-test:
     - sleep 30
     - ./overhead-test/provision/wait-for-test-complete.sh
     - ./overhead-test/provision/fetch-results.sh
-    - ./overhead-test/provision/create-results-pr.sh
\ No newline at end of file
+    - ./overhead-test/provision/create-results-pr.sh
+
+oss-scan:
+  stage: verify
```
pellared, line 187: How about moving it to the `build` stage so that it executes sooner and in parallel?

pellared (approved):

.gitlab-ci.yml
```diff
@@ -177,4 +181,10 @@ run-overhead-test:
...
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "main"'
```
pellared, line 189: Who will trigger the scans? Maybe we should use a cron schedule? 

Reference: https://docs.gitlab.com/ee/ci/pipelines/schedules.html#add-a-pipeline-schedule

.gitlab-ci.yml
pjanotti: I didn't want to run the scan on broken builds. That said as we get more familiar with FOSSA we may revisit this.

PR #640, opened 2022-10-12
[AlwaysOnProfiler] handle memory profiling configuration on managed side

lachmatt:
## Why

Handle memory profiling configuration on the managed side.

## What

- Attempt to consume and export allocation samples only if memory profiling is enabled
- Rename properties related to AlwaysOnProfiling
- Reorganize the code related to exporting cpu/allocation samples
- Update `internal-config.md`

## Tests

Existing tests.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "allowPrerelease": true
+    "allowPrerelease": false
   }
 }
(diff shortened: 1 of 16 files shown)


global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "allowPrerelease": true
+    "allowPrerelease": false
```
lachmatt, line 3: there are problems when building with prerelease version of the sdk

dszmigielski (approved):
LGTM

Kielek (approved):

PR #643, opened 2022-10-17
[Runtime Metrics] Time in gc metric

lachmatt:
## Why

Reintroduce metric that shows time spent in gc pause.

## What

- reintroduce implementation from upstream (.NET Core only)
- send as a double-valued counter
- move pushing gc collection counts metrics to common place

For now, I extended existing metrics interface. We should consider using metric sender directly (requires additional changes, postponed for now).

## Tests

Included in PR.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -18,15 +18,16 @@ To learn about differences between SignalFx metric types, visit [documentation](
 The names and the metric structure of the metrics exported are aligned with OpenTelemetry
 [implementation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/bc947a00c3f859cc436f050e81172fc1f8bc09d7/src/OpenTelemetry.Instrumentation.Runtime).
 
-| Metric                                                  | Description                                                                                                                      | Type               |
-|:--------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------|:-------------------|
-| `process.runtime.dotnet.exceptions.count`               | The count of exceptions that have been thrown in managed code since the previous observation.                                    | Counter            |
-| `process.runtime.dotnet.gc.collections.count`           | The number of garbage collections that have occurred since the process was started.                                              | CumulativeCounter  |
-| `process.runtime.dotnet.gc.heap.size`                   | The heap size, as observed during the latest garbage collection.                                                                 | Gauge              |
-| `process.runtime.dotnet.gc.allocations.size`            | The count of bytes allocated on the managed GC heap since the process was started. (.NET Core only)                              | CumulativeCounter  |
-| `process.runtime.dotnet.gc.committed_memory.size`       | The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. (.NET 6 only)  | Gauge              |
-| `process.runtime.dotnet.monitor.lock_contention.count`  | The number of times there was contention when trying to acquire a monitor lock since the process was started.                    | CumulativeCounter  |
-| `process.runtime.dotnet.thread_pool.threads.count`      | The number of thread pool threads, as observed during the latest measurement. (.NET Core only)                                   | Gauge              |
+| Metric                                                 | Description                                                                                                                     | Type              |
+|:-------------------------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------|:------------------|
+| `process.runtime.dotnet.exceptions.count`              | The count of exceptions that have been thrown in managed code since the previous observation.                                   | Counter           |
+| `process.runtime.dotnet.gc.collections.count`          | The number of garbage collections that have occurred since the process was started.                                             | CumulativeCounter |
+| `process.runtime.dotnet.gc.heap.size`                  | The heap size, as observed during the latest garbage collection.                                                                | Gauge             |
+| `process.runtime.dotnet.gc.allocations.size`           | The count of bytes allocated on the managed GC heap since the process was started. (.NET Core only)                             | CumulativeCounter |
+| `process.runtime.dotnet.gc.committed_memory.size`      | The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. (.NET 6 only) | Gauge             |
+| `process.runtime.dotnet.gc.pause.time`                 | The number of milliseconds spent in GC pause. (.NET Core only)                                                                  | Counter           |
+| `process.runtime.dotnet.monitor.lock_contention.count` | The number of times there was contention when trying to acquire a monitor lock since the process was started.                   | CumulativeCounter |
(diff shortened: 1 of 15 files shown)


Kielek (approved):

docs/metrics.md
```diff
@@ -18,15 +18,16 @@ To learn about differences between SignalFx metric types, visit [documentation](
...
+| `process.runtime.dotnet.gc.allocations.size`           | The count of bytes allocated on the managed GC heap since the process was started. (.NET Core only)                             | CumulativeCounter |
+| `process.runtime.dotnet.gc.committed_memory.size`      | The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. (.NET 6 only) | Gauge             |
+| `process.runtime.dotnet.gc.pause.time`                 | The number of milliseconds spent in GC pause. (.NET Core only)                                                                  | Counter           |
+| `process.runtime.dotnet.monitor.lock_contention.count` | The number of times there was contention when trying to acquire a monitor lock since the process was started.                   | CumulativeCounter |
+| `process.runtime.dotnet.thread_pool.threads.count`     | The number of thread pool threads, as observed during the latest measurement. (.NET Core only)                                  | Gauge             |
```
theletterf, line 30:
```suggestion
|:-------------------------------------------------------|:--------------------------------------------------------------------------------------------------------------------------------|:------------------|
| `process.runtime.dotnet.exceptions.count`              | Count of exceptions since the previous observation.                                   | Counter           |
| `process.runtime.dotnet.gc.collections.count`          | Number of garbage collections since the process started.                                             | CumulativeCounter |
| `process.runtime.dotnet.gc.heap.size`                  | Heap size, as observed during the last garbage collection.                                                                | Gauge             |
| `process.runtime.dotnet.gc.allocations.size`           | Count of bytes allocated on the managed GC heap since the process started. (.NET Core only)                             | CumulativeCounter |
| `process.runtime.dotnet.gc.committed_memory.size`      | Amount of committed virtual memory for the managed GC heap, as observed during the last garbage collection. (.NET 6 only) | Gauge             |
| `process.runtime.dotnet.gc.pause.time`                 | Number of milliseconds spent in GC pause. (.NET Core only)                                                                  | Counter           |
| `process.runtime.dotnet.monitor.lock_contention.count` | Contentions count when trying to acquire a monitor lock since the process started.                   | CumulativeCounter |
| `process.runtime.dotnet.thread_pool.threads.count`     | Number of thread pool threads, as observed during the last measurement. Only available for .NET Core.                                  | Gauge             |
```

theletterf (changes requested):

theletterf (approved):

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
```diff
@@ -172,6 +173,19 @@ private void HandleEvent(EventWrittenEventArgs eventData)
             {
                 ExtractCounters(eventData.Payload);
             }
+            else if (eventData.EventId == EventGcSuspendBegin)
+            {
+                _gcStart = eventData.TimeStamp;
+            }
+            else if (eventData.EventId == EventGcRestartEnd)
+            {
+                var start = _gcStart;
```
pjanotti, line 182: It seems safer to set `_gcStart` to null after saving the current value to the local. At least with ETW it was possible to miss some events - I'm not sure if that is the case for EventPipe but the cost is very low to avoid bogus data.

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
```diff
@@ -27,6 +27,8 @@ internal class RuntimeEventListener : EventListener, IRuntimeMetricsListener
         private const int EventContentionStop = 91;
 
         private const int EventGcHeapStats = 4;
+        private const int EventGcSuspendBegin = 9;
```
pjanotti, line 30: This got me on the wrong track: the MS doc seems to be wrong putting the event ID as 8, see https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-garbage-collection-events#gcsuspendeebegin_v1-event

At least I think this is the intended event. That said I'm now questioning if start with id 9 shouldn't be matched with end id 8 instead. I assume this is based on the OTel SDK, could you please link the corresponding code/PR?

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
```diff
@@ -172,6 +173,19 @@ private void HandleEvent(EventWrittenEventArgs eventData)
             {
                 ExtractCounters(eventData.Payload);
             }
+            else if (eventData.EventId == EventGcSuspendBegin)
```
pjanotti, line 176: It would be nice to have link here or on the enum to some doc describing these events.

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
```diff
@@ -172,6 +173,19 @@ private void HandleEvent(EventWrittenEventArgs eventData)
...
+            {
+                _gcStart = eventData.TimeStamp;
+            }
+            else if (eventData.EventId == EventGcRestartEnd)
+            {
+                var start = _gcStart;
+
+                if (start != null)
+                {
+                    _statsd.IncrementDouble(MetricsNames.Gc.PauseTime, (eventData.TimeStamp - start.Value).TotalMilliseconds);
```
pjanotti, line 186: What are the dimensions/tags that are attached to this metric?

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/643/commits/8386277b2b543869362f412e61a30a706f00c912

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
lachmatt: fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/643/commits/8386277b2b543869362f412e61a30a706f00c912

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs
lachmatt: default tags from https://github.com/signalfx/signalfx-dotnet-tracing/blob/ba64bece9ad776f361abe593e5f3d449cdeaadd5/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L199

pjanotti (approved):

PR #650, opened 2022-10-28
[Runtime Metrics] Enable runtime metrics with memory profiling

lachmatt:
## Why

Enable runtime metrics when memory profiling is enabled.

## What

- consider memory profiling configuration when enabling runtime metrics
- update `metrics.md`

## Tests

Included in PR.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -10,6 +10,8 @@ to Splunk Observability Cloud.
 To enable runtime metrics, set the `SIGNALFX_RUNTIME_METRICS_ENABLED` environment
 variable to `true` for your .NET process.
 
+> Runtime metrics are automatically enabled when [AlwaysOn memory profiling](internal/memory-profiling.md) is enabled. 
+
 The following metrics are collected by default after enabling .NET metrics.
 To learn about differences between SignalFx metric types, visit [documentation](https://docs.splunk.com/Observability/metrics-and-metadata/metric-types.html#metric-types).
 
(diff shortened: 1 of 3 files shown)


tracer/test/Datadog.Trace.Tests/TracerSettingsTests.cs
```diff
@@ -106,6 +106,24 @@ public void ReplaceLocalhost(string original, string expected)
             Assert.Equal(expected, tracerSettings.ExporterSettings.AgentUri.ToString());
         }
 
+        [Theory]
+        [InlineData("1", "0", true)]
```
Kielek, line 110: I think that you have missed `null` as correct value for the test cases.
I think that, null, 1  => true, but "0", "1" => false.

docs/metrics.md
```diff
@@ -10,6 +10,8 @@ to Splunk Observability Cloud.
 To enable runtime metrics, set the `SIGNALFX_RUNTIME_METRICS_ENABLED` environment
 variable to `true` for your .NET process.
 
+> Runtime metrics are automatically enabled when [AlwaysOn memory profiling](internal/memory-profiling.md) is enabled.
```
theletterf, line 13: What happens if you set `SIGNALFX_RUNTIME_METRICS_ENABLED` to `false` and enable memory profiling? Unlikely, but possible. Profiling settings overrides Runtime metrics? We might want to add this.

theletterf (approved):

docs/metrics.md
lachmatt: > Profiling settings overrides Runtime metrics

This is expected behavior as far as I know. Updated the docs.

tracer/test/Datadog.Trace.Tests/TracerSettingsTests.cs
lachmatt: Added additional test cases. Current implementation seems to be aligned with other languages and is expected as far as I know. I will confirm that.

Kielek (approved):

PR #655, opened 2022-11-03
Add .NET 7 to CI

Kielek:
## Why

Have test coverage on .NET 7

## What

 .NET 7 in CI

## Tests

CI


tracer/build/_build/Build.Steps.cs
```diff
@@ -1004,40 +1004,41 @@ void ExtractDebugInfoAndStripSymbols()
 
             try
             {
-                DotNetTest(config => config
-                    .SetDotnetPath(TargetPlatform)
+                 DotNetMSBuild(s => s
```
dszmigielski, line 1007: What is a purpose of this change?

Kielek: There is no possibility (or I cannot find it) to successfully pass TargetPlatform through `dotnet test` command to the `dotnet msbuild`.

.NET 7 changed path where it set output, so this parameter has to be passed.

 I decided to convert `dotnet test` to `dotnet msbuild` (which was triggered under the hood) with correct parameters.

dszmigielski (approved):

tracer/build/_build/Build.Steps.cs
pellared: Maybe you could create (or find) an issue in https://github.com/dotnet/sdk? 

I think it can be a problem not only for us. AFAIK `dotnet` CLI was created in order to reduce the usage of `msbuild`.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -29,7 +29,11 @@ public AwsSqsTests(ITestOutputHelper output)
             SetEnvironmentVariable("SIGNALFX_PROPAGATORS", "B3");
         }
 
+#if NET7_0
+        [SkippableTheory(Skip = "APMI-3564 - fix .NET 7 tests")]
```
pellared, line 33: AFAIK you could simply do:

```csharp
[Theory(Skip = "APMI-3564 - fix .NET 7 tests")]
```

😉 

Applies to other places as well.

tracer/test/Datadog.Trace.Tests/ActivityTests.cs
```diff
@@ -11,7 +11,7 @@
 using Datadog.Trace.TestHelpers;
 using FluentAssertions;
 using Xunit;
-using sd = System.Diagnostics;
+using Sd = System.Diagnostics;
```
pellared, line 14: nit: I would prefer simply using `System.Diagnostics.` intead of the `Sd` alias. It would be more readable.

pellared (approved):
only nit comments

tracer/test/Datadog.Trace.Tests/ActivityTests.cs
Kielek: I still trying to minimize changes in case of upstream sync...

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
Kielek: I expect that this construction will be removed shortly. I will keep it as it is.

tracer/build/_build/Build.Steps.cs
Kielek: I will put on my internal to do list.

PR #657, opened 2022-11-09
Update .NET7.0 to stable version

Kielek:
## Why

MS just released stable, we can drop testing with RC.

## What

Update .NET7.0 to stable version

## Tests

CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,8 +41,7 @@ on:
 
 env:
   buildConfiguration: Release
-  dotnetSdkVersion: 7.0.100-rc.2.22477.23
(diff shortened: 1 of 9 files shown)


dszmigielski (approved):

.github/workflows/ci.yml
```diff
@@ -78,7 +77,7 @@ jobs:
           3.1.424
           5.0.408
           6.0.402
```
pellared, line 79: they were also some patch releases AFAIK. may be worth to update as well

pellared (approved):

.github/workflows/ci.yml
Kielek: 08e3c86435b1442de1f176b2a6168d491b3c8d24

PR #658, opened 2022-11-10
[AlwaysOnProfiler] enable stack snapshots when only allocation profiling is enabled

lachmatt:
## Why

Enable stack snapshots when only allocation profiling is enabled.

## What

Set required flags to be able to do stack snapshots when only allocation profiling is enabled. 
DoStackSnapshot returns [CORPROF_E_INCONSISTENT_WITH_FLAGS](https://github.com/dotnet/runtime/blob/db213657acc53fc48212867d5728e83d9e36a558/src/coreclr/inc/corerror.xml#L1091) which results in empty stack traces when additional flags are not set.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -175,6 +175,21 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
 
     const auto metThreadSamplingRequirements = info10 != nullptr;
 
+    // get ICorProfilerInfo12 for >= .NET 5.0
+    ICorProfilerInfo12* info12 = nullptr;
+    hr = info_->QueryInterface(__uuidof(ICorProfilerInfo12), (void**) &info12);
+
+    if (SUCCEEDED(hr))
+    {
+        Logger::Debug("Interface ICorProfilerInfo12 found.");
+    }
+    else
+    {
+        info12 = nullptr;
+    }
+
+    const auto metAllocationSamplingRequirements = info12 != nullptr;
+
     auto pInfo = info10 != nullptr ? info10 : this->info_;
     auto work_offloader = std::make_shared<RejitWorkOffloader>(pInfo);
 
@@ -188,7 +203,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
                        COR_PRF_MONITOR_MODULE_LOADS | COR_PRF_MONITOR_ASSEMBLY_LOADS | COR_PRF_MONITOR_APPDOMAIN_LOADS |
                        COR_PRF_ENABLE_REJIT;
 
-    if (IsThreadSamplingEnabled() && metThreadSamplingRequirements)
+    if (IsThreadSamplingEnabled() && metThreadSamplingRequirements || IsAllocationSamplingEnabled() && metAllocationSamplingRequirements)
     {
         event_mask |= COR_PRF_MONITOR_THREADS | COR_PRF_ENABLE_STACK_SNAPSHOT;
     }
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -188,7 +203,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
                        COR_PRF_MONITOR_MODULE_LOADS | COR_PRF_MONITOR_ASSEMBLY_LOADS | COR_PRF_MONITOR_APPDOMAIN_LOADS |
                        COR_PRF_ENABLE_REJIT;
 
-    if (IsThreadSamplingEnabled() && metThreadSamplingRequirements)
+    if (IsThreadSamplingEnabled() && metThreadSamplingRequirements || IsAllocationSamplingEnabled() && metAllocationSamplingRequirements)
```
Kielek, line 206: I hope the hierarchy of applying operators is correct,
but I would like to see some additional parentheses here
```
    if ((IsThreadSamplingEnabled() && metThreadSamplingRequirements) || (IsAllocationSamplingEnabled() && metAllocationSamplingRequirements))
```

lachmatt: That's what I started with, vs marked them as redundant, added them back in https://github.com/signalfx/signalfx-dotnet-tracing/pull/658/commits/d70014a005865c21812e99b626533415962bacfd

Kielek (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -175,6 +175,21 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
+    hr = info_->QueryInterface(__uuidof(ICorProfilerInfo12), (void**) &info12);
+
+    if (SUCCEEDED(hr))
+    {
+        Logger::Debug("Interface ICorProfilerInfo12 found.");
+    }
+    else
+    {
+        info12 = nullptr;
+    }
```
pellared, line 189: Is this neeeded?

pellared (approved):

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -175,6 +175,21 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
 
     const auto metThreadSamplingRequirements = info10 != nullptr;
 
+    // get ICorProfilerInfo12 for >= .NET 5.0
+    ICorProfilerInfo12* info12 = nullptr;
+    hr = info_->QueryInterface(__uuidof(ICorProfilerInfo12), (void**) &info12);
+
```
pellared, line 181: nit: I would remove this blank line

1. consistency
2. to show that the condition it is tightly coupled to the previous call

lachmatt: removed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/658/commits/baa20583043c0400e8e808e8596f03fe5f554812

tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
lachmatt: This is consistent with [above](https://github.com/signalfx/signalfx-dotnet-tracing/blob/d70014a005865c21812e99b626533415962bacfd/tracer/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L173)

pjanotti (commented):
LGTM

PR #668, opened 2022-11-17
Separate thread sampling and export intervals

dszmigielski:
## Why

There is one variable for gathering thread samples interval and exporter interval. This can be quite misleading when using only memory profiling.

## What

Introduce new `SIGNALFX_PROFILER_EXPORT_INTERVAL` variable that is responsible for exporter interval only.

### DIFF
--- a/docs/advanced-config.md
+++ b/docs/advanced-config.md
@@ -158,7 +158,8 @@ of SignalFx Instrumentation for .NET.
 |-|-|-|
 | `SIGNALFX_PROFILER_LOGS_ENDPOINT` | The URL to where logs are exported using [OTLP/HTTP v1 log protocol](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md) | `http://localhost:4318/v1/logs` |
 | `SIGNALFX_PROFILER_ENABLED` | Enable to activate thread sampling. | `false` |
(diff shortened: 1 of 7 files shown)


pellared:
Have you checked how Java and JS are doing it? I would follow the same practice. Take notice that the profiler env vars are defined here: https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries

dszmigielski:
> Have you checked how Java and JS are doing it? I would follow the same practice. Take notice that the profiler env vars are defined here: https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries

Our specification does not go into such details on the configuration variables.

dszmigielski:
I wasn’t able to find something similar in Java. The previous config variable is not in the spec either, but we need it anyway. It is confusing as its description does not indicate it influences the exporting too.

lachmatt:
> > May I add this to the official docs already?
> 
> Not before it is merged (and released). Personally, I am hesitant if we should include this config... @lachmatt do you have an opinion?

I think this is a case of what was noted [here](https://github.com/signalfx/gdi-specification/issues/198#issuecomment-1313843164) - something that makes sense for this specific instrumentation.

The previous config is in spec, but with different [prefix](https://github.com/signalfx/gdi-specification/blob/main/specification/configuration.md#instrumentation-libraries).

I'd prefer to go with what's in this PR as this reduces confusion, at the same time I see @pellared point.

PR #677, opened 2022-11-23
Enable metrics by type

dszmigielski:
## Why

We want to give users granular choice of which metrics they want to be provided with.

## What

Add new setting `SIGNALFX_METRICS_{0}_ENABLED` that allows the user to pick exact groups of metrics that will be sent.

## Tests

Old and new unit tests, fixed integration tests.


tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -198,12 +198,6 @@ internal static partial class ConfigurationKeys
         /// </summary>
         public const string TracerMetricsEnabled = "SIGNALFX_TRACE_METRICS_ENABLED";
 
-        /// <summary>
-        /// Configuration key for enabling or disabling runtime metrics sent to DogStatsD.
-        /// Default value is <c>false</c> (disabled).
-        /// </summary>
-        public const string RuntimeMetricsEnabled = "SIGNALFX_RUNTIME_METRICS_ENABLED";
```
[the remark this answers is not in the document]
dszmigielski: Yes, I haven't done the docs yet

theletterf (approved):

docs/metrics.md
```diff
@@ -51,13 +51,13 @@ Names and metric structure of the metrics exported are aligned with the OpenTele
...
-| `runtime.dotnet.aspnetcore.requests.current`            | The current number of HTTP requests that have started, but not yet stopped. (.NET Core only)                                      | Gauge              |
-| `runtime.dotnet.aspnetcore.requests.failed`             | The number of failed HTTP requests received by the server. (.NET Core only)                                                       | Gauge              |
-| `runtime.dotnet.aspnetcore.requests.queue_length`       | The current length of the HTTP request queue.                                                                                     | Gauge              |
-| `runtime.dotnet.aspnetcore.requests.total`              | The total number of HTTP requests received by the server. (.NET Core only)                                                        | Gauge              |
+| `signalfx.dotnet.aspnetcore.connections.current`         | The current number of active HTTP connections to the web server. (.NET Core only)                                                 | Gauge              |
```
Kielek, line 54: why signalfx? We should keep the naming from OTel repo.

docs/metrics.md
```diff
@@ -48,13 +51,13 @@ Names and metric structure of the metrics exported are aligned with the OpenTele
...
```
pellared, line 54: Does .NET core only means that it is not supported on .NET 5.0, .NET 6.0, and .NET 7.0?
Does it mean that it is supported ONLY on .NET Core 3.1?

docs/metrics.md
lachmatt: Implementation of this metrics category is not adjusted to OTel yet.

docs/metrics.md
dszmigielski: @lachmatt ?

lachmatt: > Does ".NET Core only" means that it is not supported on .NET 5.0, .NET 6.0, and .NET 7.0? Does it mean that it is supported ONLY on .NET Core 3.1?

There metrics are available on targets listed.

pellared: @dszmigielski Can you please remove ` (.NET Core only)` from all the rows as it is working also on .NET 5.0, .NET 6.0, .NET 7.0 and it is already in an ASP.NET Core section?

docs/advanced-config.md
```diff
@@ -58,7 +58,7 @@ The following settings are common to most instrumentation scenarios:
 | `SIGNALFX_PROFILER_PROCESSES` | Sets the filename of executables the profiler can attach to. Supports multiple semicolon-separated values, for example: `MyApp.exe;dotnet.exe` |  |
 | `SIGNALFX_TRACE_CONFIG_FILE` | Path of the JSON configuration file. |  |
 | `SIGNALFX_TRACE_ENABLED` | Enable to activate the tracer. | `true` |
-| `SIGNALFX_RUNTIME_METRICS_ENABLED` | Enable to activate the runtime metrics. | `false` |
+| `SIGNALFX_METRICS_{0}_ENABLED` | Configuration pattern for enabling or disabling a specific metrics group. For example, in order to enable `NetRuntime` metrics, set `SIGNALFX_METRICS_NetRuntime_ENABLED=true`. Supported metrics are `NetRuntime`, `Process`, `AspNet`. | `false` |
```
pellared, line 61: I think we should refer to `metrics.md`. And in `metrics.md` we should list the supported IDs. Right now it is nowhere explicitly said that `AspNet` is for "ASP.NET Core metrics"

tracer/src/Datadog.Trace/Configuration/MetricsIntegrationId.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Datadog.Trace.Configuration;
+
+internal enum MetricsIntegrationId
+{
+    AspNet,
```
pellared, line 5: How about `AspNetCore`?

dszmigielski: I don't mind changing that. Anyone against?

docs/metrics.md
dszmigielski: Sure

tracer/src/Datadog.Trace/Configuration/MetricsIntegrationId.cs
dszmigielski: Done

pellared (approved):
LGTM but it would be good to have one more reviewer (the PR is big)

dev/envvars.sh
```diff
@@ -55,4 +55,4 @@ export SIGNALFX_TRACE_DEBUG="1"
 export SIGNALFX_DUMP_ILREWRITE_ENABLED="0"
 export SIGNALFX_CLR_ENABLE_INLINING="1"
 export SIGNALFX_ENV=${SIGNALFX_ENV:-$(whoami)}
-export SIGNALFX_RUNTIME_METRICS_ENABLED="1"
+export SIGNALFX_METRICS_NetRuntime_ENABLED="1"
```
RassK, line 58: Can we use only uppercase? having exact case requirement seems bad UX and very error prone.

Kielek: I think that the same behavior we have for traces. It will be great to support both...
It can be separate issue.

tracer/src/Datadog.Trace/Configuration/ImmutableMetricsIntegrationSettings.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace Datadog.Trace.Configuration;
```
Kielek, line 1: Please check all files and add `// Modified by Splunk Inc.` where needed.
If it is newly created file, please just add this header also.

dev/envvars.sh
dszmigielski: Yep, this follows the same rules as traces env vars. I would rather not change that in this PR

Kielek (approved):

PR #682, opened 2022-11-28
[RuntimeMetrics] Apply fix from upstream

lachmatt:
## Why

Apply fix from upstream


tracer/src/Datadog.Trace/Vendors/StatsdClient/Worker/Waiter.cs
```diff
@@ -3,15 +3,15 @@
 // This file was automatically generated by the UpdateVendors tool.
 //------------------------------------------------------------------------------
 using System;
-using System.Threading.Tasks;
+using System.Threading;
 
 namespace Datadog.Trace.Vendors.StatsdClient.Worker
 {
     internal class Waiter : IWaiter
```
dszmigielski, line 10: image

dszmigielski (approved):

Kielek (approved):

PR #687, opened 2022-12-02
Apply consistent set of attributes to metrics and stacks sent in logs

lachmatt:
## Why

Have consistent set of attributes across different signals.
## What

- add attributes commonly added by Otel-based instrumentations to metrics/stacks sent in logs
- rename `SIGNALFX_TRACE_GLOBAL_TAGS` to `SIGNALFX_GLOBAL_TAGS` as it applies broader (spans and stack sent in logs before, now also to metrics)
- replace `Verify` usage with standard assertions for simple scenario in `AlwaysOnProfilerTests`
- replace `Xunit` assertions with `FluentAssertions` in `RuntimeMetricsTests`

## Tests

Included in PR.


Kielek (approved):

theletterf:
@lachmatt Can this be documented without documenting the runtime / profiling bits?

docs/advanced-config.md
```diff
@@ -66,7 +66,8 @@ The following settings are common to most instrumentation scenarios:
 |-|-|-|
 | `SIGNALFX_DISABLED_INTEGRATIONS` | Comma-separated list of disabled library instrumentations. The available integration ID values can be found [here](instrumented-libraries.md). |  |
 | `SIGNALFX_RECORDED_VALUE_MAX_LENGTH` | The maximum length an attribute value can have. Values longer than this are truncated. Values are discarded entirely when set to 0, and ignored when set to a negative value. | `12000` |
-| `SIGNALFX_TRACE_GLOBAL_TAGS` | Comma-separated list of key-value pairs that specify global span tags. For example: `"key1:val1,key2:val2"` |  |
+| `SIGNALFX_TRACE_GLOBAL_TAGS` | Deprecated. Use `SIGNALFX_GLOBAL_TAGS` instead. |  |
+| `SIGNALFX_GLOBAL_TAGS` | Comma-separated list of key-value pairs that specify global tags added to all telemetry signals. For example: `"key1:val1,key2:val2"`. If not set, value of `SIGNALFX_TRACE_GLOBAL_TAGS` is used if set. |  |
```
theletterf, line 70:
```suggestion
| `SIGNALFX_GLOBAL_TAGS` | Comma-separated list of key-value pairs that specify global tags added to all telemetry signals. For example: `"key1:val1,key2:val2"`. If unset, the value of the `SIGNALFX_TRACE_GLOBAL_TAGS` environment variable is used, if available. |  |
```

What happens if `SIGNALFX_TRACE_GLOBAL_TAGS` is also unset?

theletterf (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -94,6 +94,7 @@ public TracerSettings(IConfigurationSource source)
                                           100;
 
             GlobalTags = source?.GetDictionary(ConfigurationKeys.GlobalTags) ??
+                         source?.GetDictionary("SIGNALFX_TRACE_GLOBAL_TAGS") ??
```
pjanotti, line 97: nit: add a comment about this still be here to help transition to the new env var

pjanotti: On the file below we have: `// backwards compatibility for names used in the past`

pjanotti (approved):
LGTM. A tiny suggestion about adding a comment.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lachmatt, line 97:
```suggestion
                         // backwards compatibility for names used in the past
                         source?.GetDictionary("SIGNALFX_TRACE_GLOBAL_TAGS") ??
```

lachmatt: added in https://github.com/signalfx/signalfx-dotnet-tracing/pull/687/commits/bac99447db1db4d758c6faab0f06f9eee4853042

PR #688, opened 2022-12-08
Remove NativeId from CPU profiler

pellared:
## Why

Fix deadlock in AlwaysOn Profiler.

## What

Remove NativeId from CPU profiler.

## Tests

- CI
- long running test (working stable since for more than 2 days)

TODO:
- manual e2e test


pellared:
@johnbley PTAL

tracer/src/Datadog.Trace/AlwaysOnProfiler/PlainTextThreadSampleExporter.cs
```diff
@@ -84,11 +84,10 @@ private static string BuildStackTrace(ThreadSample threadSample)
 
             stackTraceBuilder.AppendFormat(
                 CultureInfo.InvariantCulture,
-                "\"{0}\" #{1} prio=0 os_prio=0 cpu=0 elapsed=0 tid=0x{2:x} nid=0x{3:x}\n",
+                "\"{0}\" #{1} prio=0 os_prio=0 cpu=0 elapsed=0 tid=0x{2:x}\n",
```
pjanotti, line 87: Given that we are using pprof by default this shouldn't be an issue but per GDI spec the non-available nid should be zero like the other fields. If that is not the case anymore for profiling the backend we should update the [GDI spec](https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md#call-stack-format-for-text-data-format) or perhaps even remove the textual representation altogether.

pjanotti (approved):
LGTM

pellared:
> I'd probably increment `kCurrentThreadSamplesBufferVersion` to `2` in `always_on_profiler.cpp` and adjust all the fallout from that (parsing code, possibly unit tests, etc.). No backwards compatibility (support for parsing v1) is required on the parser since these binary components are only meant to work in version lockstep, but I added the "protocol" version number just as a safety precaution.

I do not feel it is worth spending the time. Especially because fixing/recreation of binary files used for parser unit test consumes some time. As long as it is internal and native+managed are shipped together I think we can keep using `1`.

Kielek (approved):

Kielek:
> I'd probably increment `kCurrentThreadSamplesBufferVersion` to `2` in `always_on_profiler.cpp` and adjust all the fallout from that (parsing code, possibly unit tests, etc.). No backwards compatibility (support for parsing v1) is required on the parser since these binary components are only meant to work in version lockstep, but I added the "protocol" version number just as a safety precaution.

For now, it is purely internal implementation. IMO we can keep it still as 1. If it could we handled as in Java - by file writing, then we should upgrade the version.

lachmatt (approved):

PR #704, opened 2023-03-17
Add deprecation notice to process.cpu.utilization

lachmatt:
## What

Deprecate `process.cpu.utilization`.


Kielek (approved):

theletterf (approved):

theletterf:
What should users use instead?

PR #709, opened 2023-03-28
[AlwaysOnProfiling] add total frame count to LogRecord attributes

lachmatt:
## Why

Required by the GDI spec.

## What

Add `profiling.data.total.frame.count` attribute to `LogRecords` containing `pprof` profiles.

## Tests

Included in PR.

### DIFF
--- a/tracer/src/Datadog.Trace/AlwaysOnProfiler/PprofThreadSampleExporter.cs
+++ b/tracer/src/Datadog.Trace/AlwaysOnProfiler/PprofThreadSampleExporter.cs
@@ -4,14 +4,17 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
+using System.Linq;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.Vendors.ProtoBuf;
+using Datadog.Tracer.OpenTelemetry.Proto.Common.V1;
 using Datadog.Tracer.Pprof.Proto.Profile;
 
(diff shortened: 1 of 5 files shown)


tracer/src/Datadog.Trace/AlwaysOnProfiler/PprofThreadSampleExporter.cs
```diff
@@ -23,13 +26,30 @@ public PprofThreadSampleExporter(ImmutableTracerSettings tracerSettings, ILogSen
...
+
+            AddLogRecord(
+                allocationProfile,
+                ProfilingDataTypeAllocation,
+                totalFrameCount);
+        }
+
+        private static int CountFrames(IEnumerable<ThreadSample> samples)
+        {
+            return samples
```
Kielek, line 51: LINQ, it is probably not the most efficient way to do it.

Kielek (approved):

tracer/src/Datadog.Trace/AlwaysOnProfiler/PprofThreadSampleExporter.cs
lachmatt: Based on my benchmarks, Linq usage would incur additional 450ns and allocate additional 32B, which I assumed is acceptable for operation that happens by default once in 10s (once in 1s max).
Fixed in https://github.com/signalfx/signalfx-dotnet-tracing/pull/709/commits/df5cb9197b32a357f9b8da86b00dc6f0fc4bc0b1

PR #715, opened 2023-04-19
Release 1.1.0

lachmatt:
### General

- Add `profiling.data.total.frame.count` attribute when exporting profiling data.


Kielek (changes requested):
Lets sync with @akubik-splunk before merge.

Kielek (approved):

PR #730, opened 2023-06-15
Replace SFx images with Splunk

pjanotti:
## Why

This came up during the OSS internal process for this project

## What

Replace SignalFx visible to customers when installing the product to respective Splunk images.

## Tests

Will manually validate the MSI after it is produced by CI.


pjanotti:
Note that tracer/src/WindowsInstaller/sfx-icon.ico was actually not used by the tracer installer. There was a reference to a file with same name in the shared installer but the file was not available there and we don't use that installer. Opted to keep the updated file under tracer since the art was already updated.

pjanotti:
msi-00-welcome

msi-01-license
 
msi-02-folder

msi-03-install

msi-04-finish

RassK (approved):

pellared (approved):

PR #740, opened 2023-10-02
Update dotnet-install.sh sha256sum

pjanotti:
## Why

Checksum is failing due to changes in the script.

## What

Updated the sha256 for the script.

## Tests

N/A


pjanotti:
It seems that we exceed some size limits for CI.

pjanotti:
Fixed via #742

PR #748, opened 2023-12-13
Fix CI

pjanotti:
## Why

Fixing CI build after .NET 8 was added to runners.

## What

* Ignore warnings NU1902 and NU1903 - raised by .NET 8 for .NET Core app 2.1
* Update dotnet-install.sh SHA256

## Tests

N/A


.github/workflows/ci.yml
```diff
@@ -80,6 +80,8 @@ jobs:
           5.0.408
           6.0.404
           7.0.101
+          8.0.100
```
Kielek, line 83: Together with these change we should probably bump all other .NET sdk versions. Not a blocker to this PR.

.nuke/build.schema.json
Kielek, line 1: Add this folder to .gitignore and remove all files from PR.

lachmatt: I think this directory is usually added as in this PR when `nuke` is used. Relevant part from the docs: https://nuke.build/faq/#build-structure

.github/workflows/ci.yml
pjanotti: Good point, let's do it separately.

.nuke/build.schema.json
pjanotti: The new version of nuke requires the `.nuke/` folder, it doesn't work with just the file `.nuke` that we had before. Given that git doesn't track empty directories, I'm going to create `.nuke/.gitkeep`.

pjanotti: Just by running the build script it also generates another file: the `.nuke/parameters.json`. In light of that I think it is better to leave it as it is.

Kielek (approved):

PR #812, opened 2024-12-10
Update actions/upload-artifact and download-artifact to v4.

breedx-splk:
Update actions/upload-artifact and download-artifact to v4. v3 is going away and will break builds in Jan.


Kielek:
Ref: https://github.com/signalfx/signalfx-dotnet-tracing?tab=readme-ov-file#warning-deprecation-notice

I hope, I will merge only one PR here, with information that it is fully deprecated.

